PR #2, opened 2018-11-30
enable/disable diskquota dynamically

gfphoenix78:


diskquota.c
```diff
@@ -244,6 +283,192 @@ disk_quota_worker_main(Datum main_arg)
...
+/**
+ * create table to record the list of monitored databases
+ */
+static void
+create_monitor_db_table()
+{
+	const char *sql;
+	sql = "create schema if not exists diskquota_catalog;"
+		"create table if not exists diskquota_catalog.database_list(dbid oid not null unique);";
+	pg_usleep(4*1000*1000);
```
zhangh43: add comment on why sleep

diskquota.c
```diff
@@ -244,6 +283,192 @@ disk_quota_worker_main(Datum main_arg)
...
+	char str[128];
+	snprintf(str, sizeof(str), "delete from diskquota_catalog.database_list where dbid=%d;", dbid);
+	exec_simple_query_(str, SPI_OK_DELETE);
+}
+static void
+try_kill_db_worker(Oid dbid)
+{
+	DiskQuotaWorkerEntry *hash_entry;
+	bool found;
+	hash_entry = (DiskQuotaWorkerEntry *)hash_search(disk_quota_worker_map,
```
zhangh43: HASH-REMOVE return a dangling pointer, better to HASH_FIND first to kill the worker and then HASH_REMOVE.

diskquota.c
gfphoenix78: In early stage, the postmaster process has not fully initialized yet. bgworker processes start before `ServerLoop`, but there are some other initializations to do in `ServerLoop`, like checkpointer, bgwriter, etc. Some aux-processes will try to open template1 or like, which is blocked by the following SPI call. So, it's better to wait for several seconds to let all initialization normal.

diskquota.c
```diff
@@ -766,3 +883,132 @@ get_size_in_mb(char *str)
...
 }
+
+/*
+ * trigger start diskquota worker when create extension diskquota
+ */
+Datum
+diskquota_start_worker(PG_FUNCTION_ARGS)
+{
+	elog(LOG, "DB = %d, MyProc=%p launcher pid=%d", MyDatabaseId, MyProc, message_box->launcher_pid);
+	elog(NOTICE, "DB = %d", MyDatabaseId);
```
zhangh43: remove this notice.
If use elog(LOG), better to add diskquota prefix and more info to let DBA what this log means

diskquota.c
```diff
@@ -766,3 +883,132 @@ get_size_in_mb(char *str)
...
+	message_box->cmd = CMD_CREATE_EXTENSION;
+	message_box->result = ERR_PENDING;
+	message_box->data[0] = MyDatabaseId;
+	// setup sig handler to receive message
+	kill(message_box->launcher_pid, SIGUSR1);
+	while(true)
+	{
+		WaitLatch(&MyProc->procLatch,
+					   WL_LATCH_SET | WL_TIMEOUT | WL_POSTMASTER_DEATH,
+					   200L, PG_WAIT_EXTENSION);
```
zhangh43: sleep when holding a exclusive lock?

zhangh43: One solution is to release the write lock before send signal to diskquota launcher, and use an 'in_use' flag in message box to identify whether this message box is currently used by a `create extension diskquota` QD(call it QD1). It will help to prevent from other QD to create diskquota extension simultaneously.
Follow this way, Diskquota launcher could request a read/write lock when access message box and QD1 could also request read lock to access message box to check whether it could continue its work. If everything is done, QD1 could request a write lock again and turn the 'in_use' flag to off.

diskquota.c
```diff
@@ -244,6 +283,192 @@ disk_quota_worker_main(Datum main_arg)
...
+static void
+create_monitor_db_table()
+{
+	const char *sql;
+	sql = "create schema if not exists diskquota_catalog;"
+		"create table if not exists diskquota_catalog.database_list(dbid oid not null unique);";
+	pg_usleep(4*1000*1000);
+	exec_simple_utility(sql);
+}
+static inline void
```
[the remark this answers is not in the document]
gfphoenix78: added it.

diskquota.c
```diff
@@ -244,6 +283,192 @@ disk_quota_worker_main(Datum main_arg)
...
+exec_simple_utility(const char *sql)
+{
+	StartTransactionCommand();
+	exec_simple_query_(sql, SPI_OK_UTILITY);
+	CommitTransactionCommand();
+}
+static void
+exec_simple_query_(const char *sql, int expected_code)
```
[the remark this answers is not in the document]
gfphoenix78: have renamed to exec_simple_spi

diskquota.c
```diff
@@ -49,12 +66,17 @@ typedef struct DiskQuotaWorkerEntry DiskQuotaWorkerEntry;
...
+	Oid dbid;
+	int pid; /* worker pid */
 	BackgroundWorkerHandle *handle;
 };
 
+MessageBox * message_box = NULL;
 /* using hash table to support incremental update the table size entry.*/
 static HTAB *disk_quota_worker_map = NULL;
+static object_access_hook_type next_object_access_hook;
+DiskQuotaLocks dq_locks;
```
zhangh43: put static together

diskquota.c
```diff
@@ -355,112 +527,22 @@ get_database_list(void)
...
-		}
-	}
 }
 
 /*
  * Dynamically launch an disk quota worker process.
  */
 static int
-start_worker(char* dbname)
+start_worker_id(Oid dbid)
```
zhangh43: naming to start_worker_by_dboid

diskquota.h
```diff
@@ -13,9 +13,43 @@ typedef struct
...
+struct DiskQuotaLocks
+{
+	LWLock *active_table_lock;
+	LWLock *black_map_lock;
+	LWLock *message_box_lock;
+};
+typedef struct DiskQuotaLocks DiskQuotaLocks;
+struct MessageBox
+{
+	int launcher_pid;
```
zhangh43: should add volatile keyword to these variables.

PR #3, opened 2018-12-05
Support in-transaction disk quota management

HaozhouWang:
1. Disk quota now support the in-transaction disk
quota usage monitor and enforcement. For example,

```
BEGIN;
CREATE TABLE tbl (i int);
insert into tbl select generate_series(1,100000000);
ABORT;
```

2. Optimized active checker logic to avoid unexpected behavior
3. Update test cases


HaozhouWang:
close it to fix confict

PR #18, opened 2019-04-02
Add test when primary segment is failed

HaozhouWang:
1. Add a new test suit to make sure diskqouta is still working
   when primary segment is failed, and mirror segment is swtich to
   primary

### DIFF
--- a/expected/test_primary_failure.out
+++ b/expected/test_primary_failure.out
@@ -1,0 +1,95 @@
+CREATE SCHEMA ftsr;
+SELECT diskquota.set_schema_quota('ftsr', '1 MB');
+ set_schema_quota 
+------------------
+ 
+(1 row)
+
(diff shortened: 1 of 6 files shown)


expected/test_primary_failure.out
```diff
@@ -0,0 +1,95 @@
...
+(2 rows)
+
+-- no sleep, it will take effect immediately
+SELECT quota_in_mb, nspsize_in_bytes from diskquota.show_schema_quota_view where schema_name='ftsr';
+ quota_in_mb | nspsize_in_bytes 
+-------------+------------------
+         200 |          1310720
+(1 row)
+
+INSERT INTO a SELECT generate_series(1,100);
```
zhangh43: why insert suceeded?

zhangh43:
+1

PR #20, opened 2019-04-04
Refactor diskquota launcher.

zhangh43:
Separate diskquota utility functions into file
diskquota_utility.c
Rename MessageBox to ExtensionDDLMessage.
Cleanup SPI and snapshot when SPI_execute error.
Add more comments for launcher code.

no new feature, so no test needed.

### DIFF
--- a/diskquota.c
+++ b/diskquota.c
@@ -10,7 +10,7 @@
  * Copyright (c) 2018-Present Pivotal Software, Inc.
  *
  * IDENTIFICATION
- *		gpcontrib/gp_diskquota/diskquota.c
+ *		diskquota/diskquota.c
  *
  * -------------------------------------------------------------------------
  */
@@ -22,7 +22,6 @@
 #include "access/xact.h"
 #include "catalog/indexing.h"
 #include "catalog/namespace.h"
-#include "catalog/objectaccess.h"
 #include "catalog/pg_collation.h"
 #include "catalog/pg_database.h"
 #include "catalog/pg_extension.h"
@@ -53,17 +52,10 @@
 #include "diskquota.h"
 PG_MODULE_MAGIC;
 
-/* disk quota helper function */
-PG_FUNCTION_INFO_V1(set_schema_quota);
-PG_FUNCTION_INFO_V1(set_role_quota);
-PG_FUNCTION_INFO_V1(diskquota_start_worker);
-PG_FUNCTION_INFO_V1(init_table_size_table);
-
-/* timeout count to wait response from launcher process, in 1/10 sec */
-#define WAIT_TIME_COUNT  1200
-
 /* max number of monitored database with diskquota enabled */
 #define MAX_NUM_MONITORED_DB 10
+
+#define DISKQUOTA_DB	"diskquota"
 
 /* flags set by signal handlers */
 static volatile sig_atomic_t got_sighup = false;
@@ -86,11 +78,10 @@ struct DiskQuotaWorkerEntry
 };
 
 DiskQuotaLocks diskquota_locks;
-MessageBox *message_box = NULL;
+ExtensionDDLMessage *extension_ddl_message = NULL;
 
 /* using hash table to support incremental update the table size entry.*/
 static HTAB *disk_quota_worker_map = NULL;
-static object_access_hook_type next_object_access_hook;
 static int	num_db = 0;
 
 /* functions of disk quota*/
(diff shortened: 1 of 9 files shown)


diskquota.c
```diff
@@ -470,69 +561,134 @@ start_workers_from_dblist()
 			elog(ERROR, "[diskquota launcher] dbid cann't be null in table database_list");
 		dbid = DatumGetObjectId(dat);
 		if (!is_valid_dbid(dbid))
```
gfphoenix78: When this happens, it's better to warn the DBA that there is something wrong with consistency.

diskquota.c
```diff
@@ -470,69 +561,134 @@ start_workers_from_dblist()
...
-		if (start_worker_by_dboid(dbid) < 1)
+		if (!start_worker_by_dboid(dbid))
 			elog(ERROR, "[diskquota launcher] start worker process of database(%u) failed", dbid);
 		num++;
+
+		/*
+		 * diskquota only supports to monitor at most MAX_NUM_MONITORED_DB
+		 * databases
+		 */
+		if (num >= MAX_NUM_MONITORED_DB)
```
gfphoenix78: We control the number of monitored databases when running `create extension diskquota`. Here it's better a debug warning. It could happen when 1. data consistency(disable diskquota and then drop database); 2. bugs in diskquota

diskquota.c
zhangh43: Add a warning in this case.
Below is a TODO to help DBA to clean the not valid dbid.

diskquota.c
zhangh43: Also add a warning message here.
Same as above we would help DBA to clean the not valid dbs

diskquota.c
```diff
@@ -345,97 +339,194 @@ disk_quota_worker_main(Datum main_arg)
...
 
-	ret = SPI_connect();
-	if (ret != SPI_OK_CONNECT)
-		elog(ERROR, "[diskquota] SPI connect error, code=%d", ret);
-	PushActiveSnapshot(GetTransactionSnapshot());
-	ret = SPI_execute(sql, false, 0);
-	if (ret != expected_code)
+		if (SPI_execute(sql, false, 0) != SPI_OK_UTILITY)
+		{
+			elog(ERROR, "[diskquota launcher] SPI_execute error, sql:'%s', code %d", sql, ret);
```
gfphoenix78: ret is not assigned to a meaningful error code

zhangh43: good catch, better to use errno to replace it.

diskquota.c
```diff
@@ -315,6 +307,8 @@ disk_quota_worker_main(Datum main_arg)
 					   diskquota_naptime * 1000L);
 		ResetLatch(&MyProc->procLatch);
```
gfphoenix78: There is some missing code to check postmaster's health, as the following code.
```
if (rc & WL_POSTMASTER_DEATH)
	proc_exit(1);
```

diskquota.c
```diff
@@ -345,97 +339,194 @@ disk_quota_worker_main(Datum main_arg)
 			/* TODO: Add hard limit function here */
 		}
 
-		/* emergency bailout if postmaster has died */
+		/* Emergency bailout if postmaster has died */
 		if (rc & WL_POSTMASTER_DEATH)
 			proc_exit(1);
```
gfphoenix78: Is it too late to check postmaster's death here? I think the launcher process should exit as soon as it knows that postmaster is dead.

diskquota.c
```diff
@@ -315,6 +307,8 @@ disk_quota_worker_main(Datum main_arg)
 					   diskquota_naptime * 1000L);
 		ResetLatch(&MyProc->procLatch);
 	}
+
+	/* Refresh quota model with init mode */
 	refresh_disk_quota_model(true);
```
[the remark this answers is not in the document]
zhangh43: should recheck got_sigterm after loop out.

diskquota.c
```diff
@@ -345,97 +339,194 @@ disk_quota_worker_main(Datum main_arg)
...
 		if (got_sighup)
 		{
 			got_sighup = false;
 			ProcessConfigFile(PGC_SIGHUP);
 		}
 	}
 
-	diskquota_invalidate_db(MyDatabaseId);
-	proc_exit(0);
+	proc_exit(1);
```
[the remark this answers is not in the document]
zhangh43: yes, should be 0

diskquota.c
zhangh43: good catch

diskquota.c
zhangh43: good catch

diskquota.c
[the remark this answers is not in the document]
gfphoenix78: Maybe it's reasonable to maintain a list of worker process, and the launcher process kills them when the launcher exits. Or, the state of diskquota is that someone sees a collection of worker processes without a launcher later.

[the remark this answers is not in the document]
gfphoenix78: Logic of SIGTERM of the launcher process: 
1.  launcher process should be still alive
2. worker processes exit, and then all blacklists are cleared.
In the end, the launcher process keeps idle, and diskquota has no effect with limitation of disk usage.

gfphoenix78:
I have reviewed the source code of spi.c two days ago. I find that there is no need to push/pop a snapshot to wrap `SPI_execute`, the function `SPI_execute` handle snapshot by themselves.

diskquota.c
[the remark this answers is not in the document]
zhangh43: Add a function terminate_all_workers() before launcher exit.

PR #24, opened 2019-04-09
Add schema & role quota quick search view

weinan003:
add view "diskquota.show_schema_quota_proximate_view" using
diskquota.table_size instead of physical stat each data file to
accelerate schema size calculation.


diskquota--1.0.sql
```diff
@@ -36,6 +36,16 @@ RETURNS void STRICT
...
 
+CREATE VIEW diskquota.show_schema_quota_proximate_view AS
+select pgns.nspname as schema_name, pgc.relnamespace as schema_oid, qc.quotalimitMB as quota_in_mb, sum(ts.size) as nspsize_in_bytes
+from diskquota.table_size as ts,
+        pg_class as pgc,
+        diskquota.quota_config as qc,
+        pg_namespace as pgns
+where ts.tableid = pgc.oid and qc.targetoid = pgc.relnamespace and pgns.oid = pgc.relnamespace
+group by relnamespace, qc.quotalimitMB, pgns.nspname
+order by pgns.nspname;
```
zhangh43: qc.targetoid = pgc.relnamespace and pgns.oid = pgc.relnamespace
pg_class seems could be delete from FromList.
Better to have fast version of show_role_quota_view

weinan003: some schema may not be a quota,  they do not need to show

PR #27, opened 2019-04-11
Import idle worker terminate mechanism

weinan003:
As normal db connection, QD terminate QE processes if the connection is
idle for while. Import this mechanism in diskquota_launcher_process

### DIFF
--- a/diskquota.c
+++ b/diskquota.c
@@ -17,6 +17,7 @@
 #include "postgres.h"
 
 #include <unistd.h>
+#include <utils/timeout.h>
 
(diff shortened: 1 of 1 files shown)


diskquota.c
```diff
@@ -432,6 +434,8 @@ disk_quota_launcher_main(Datum main_arg)
 	start_workers_from_dblist();
 
 	/* main loop: do this until the SIGTERM handler tells us to terminate. */
+    EnableClientWaitTimeoutInterrupt();
```
zhangh43: better to use pgindent to format code

PR #49, opened 2019-12-12
Concourse can generate filename based on OS type

HaozhouWang:


HaozhouWang:
Quick fix, no need to review

PR #52, opened 2020-09-11
Fix "out of shared memory" warning on hash_search() in report_active_table_helper()

### LINKED ISSUE #51: Warning: out of shared memory
## Issue
Some GPDB clusters generate warnings in common conditions (with no significant external influence present). Warnings are in form
```
WARNING:  out of shared memory
WARNING:  Share memory is not enough for active tables.
```

This is a known issue [mentioned in the README](https://github.com/greenplum-db/diskquota/tree/362fb52d24eb99240ed5d7097d8b3099f4ebff89#known-issue). The document also offers a solution: increase `diskquota.max_active_tables` GUC cluster-wide (and restart the database).

However, this solution does not work: no matter how huge the mentioned value is, the warnings do not disappear.

## Cause
I have made some research of the code and identified the probable cause of the problem.

The warning is generated at 

https://github.com/greenplum-db/diskquota/blob/362fb52d24eb99240ed5d7097d8b3099f4ebff89/gp_activetable.c#L171

and its cause is the lack of shared memory to allocate a new hash map entry. This is the same the warning message says.

The **cause** of this lack of shared memory is the wrong order of operations in [`_PG_init()`](https://github.com/greenplum-db/diskquota/blob/362fb52d24eb99240ed5d7097d8b3099f4ebff89/diskquota.c#L113). There,

https://github.com/greenplum-db/diskquota/blob/362fb52d24eb99240ed5d7097d8b3099f4ebff89/diskquota.c#L131

is called, [which](https://github.com/greenplum-db/diskquota/blob/362fb52d24eb99240ed5d7097d8b3099f4ebff89/quotamodel.c#L151) then [calls](https://github.com/greenplum-db/diskquota/blob/362fb52d24eb99240ed5d7097d8b3099f4ebff89/quotamodel.c#L164) `DiskQuotaShmemSize()` method, which uses the value of `diskquota.max_active_tables` GUC:

https://github.com/greenplum-db/diskquota/blob/362fb52d24eb99240ed5d7097d8b3099f4ebff89/quotamodel.c#L247

However, this GUC value is initialized *after* the mentioned call from `_PG_init()`:

https://github.com/greenplum-db/diskquota/blob/362fb52d24eb99240ed5d7097d8b3099f4ebff89/diskquota.c#L149

The default value of this GUC is

https://github.com/greenplum-db/diskquota/blob/362fb52d24eb99240ed5d7097d8b3099f4ebff89/diskquota.c#L70

which means no shared memory space is reserved for the active tables hash map.

As a result, the active tables hash map uses memory reserved somewhere else (e.g. by some other extension). In case of some clusters, this amount is not enough, and the warnings mentioned above are generated.

### Redefine GUC value / eliminate GUC
**The error only affects [gpdb](https://github.com/greenplum-db/diskquota/tree/gpdb) branch**. In [master](https://github.com/greenplum-db/diskquota/tree/master), the default value of the GUC was [changed to `1024 * 1024`](https://github.com/greenplum-db/diskquota/commit/025f3fa2940c8afa8691ed4c98eb0d2abc184fa4#diff-a9cc13af9d3c857ea60c351938f5b165R65) and the error does not manifest. In fact, `diskquota.max_active_tables` GUC may be eliminated and replaced with a constant, as it is not used for any other purpose (at least in the current code).

### Reposition GUC initialization
The correct (intended) behaviour can be achieved by repositioning of GUC initialization, so that it is initialized before `RequestAddinShmemSpace()` call is made.

leskin-in:
Closes #51 

GUC variables must be declared before any other actions are made by diskquota. In particular, the value of `diskquota_max_active_tables` (GUC `diskquota.max_active_tables`) is used in `DiskQuotaShmemSize()`. The late GUC variable declaration caused this variable to be `0`, thus leading to allocation of insufficient amount of memory.

Fix this:
1. Move GUC declarations to a separate static function
2. Call this function before any other actions performed in `_PG_init()`

### DIFF
--- a/diskquota.c
+++ b/diskquota.c
@@ -95,6 +95,7 @@ void		disk_quota_launcher_main(Datum);
 
 static void disk_quota_sigterm(SIGNAL_ARGS);
 static void disk_quota_sighup(SIGNAL_ARGS);
+static void define_guc_variables(void);
 static bool start_worker_by_dboid(Oid dbid);
 static void start_workers_from_dblist(void);
 static void create_monitor_db_table(void);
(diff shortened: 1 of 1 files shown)


Sasasu:
thanks for the contribution!

this commit merged into gpdb branch in PR #54 (9a62a0977ab1d90dad89e80b087ae5d548568b53)

I am closing this PR  since the commit has merged.

PR #58, opened 2021-05-24
Support tablespace-based quota

xuebinsu:


diskquota--1.0.sql
```diff
@@ -8,6 +8,14 @@ CREATE SCHEMA diskquota;
 -- Configuration table
 CREATE TABLE diskquota.quota_config (targetOid oid, quotatype int, quotalimitMB int8, PRIMARY KEY(targetOid, quotatype));
 
+CREATE TABLE diskquota.target (
+        rowId serial, --REFERENCES diskquota.quota_config.targetOid,
+        quotatype int, --REFERENCES disquota.quota_config.quotatype,
+        primaryOid oid,
+        tablespaceOid oid, --REFERENCES pg_tablespace.oid,
+        PRIMARY KEY (rowId, quotatype)
```
fanfuxiaoran, line 16: I think primary key should be (rowId, quotatype,tablespaceOid)

diskquota--1.0.sql
```diff
@@ -8,6 +8,14 @@ CREATE SCHEMA diskquota;
 -- Configuration table
 CREATE TABLE diskquota.quota_config (targetOid oid, quotatype int, quotalimitMB int8, PRIMARY KEY(targetOid, quotatype));
 
+CREATE TABLE diskquota.target (
+        rowId serial, --REFERENCES diskquota.quota_config.targetOid,
+        quotatype int, --REFERENCES disquota.quota_config.quotatype,
+        primaryOid oid,
```
fanfuxiaoran, line 14: What does the primaryOid mean?

diskquota--1.0.sql
```diff
@@ -8,6 +8,39 @@ CREATE SCHEMA diskquota;
...
+        tablespaceOid oid, --REFERENCES pg_tablespace.oid,
+        PRIMARY KEY (id, quotatype)
+);
+
+CREATE OR REPLACE FUNCTION diskquota.set_tablespace_quota(ts oid, quota int8)
+RETURNS void AS $$
+        INSERT INTO diskquota.quota_config VALUES (ts, 2, quota);
+$$ LANGUAGE SQL;
+
+CREATE OR REPLACE FUNCTION diskquota.set_namespace_tablespace_quota(ns oid, ts oid, quota int8)
```
fanfuxiaoran, line 24: I think the user would like to use the name of tablespace than OID  here.

diskquota--1.0.sql
```diff
@@ -8,6 +8,39 @@ CREATE SCHEMA diskquota;
...
+RETURNS void AS $$
+DECLARE 
+    target_id diskquota.target.id%TYPE;
+BEGIN
```
fanfuxiaoran, line 28: If we don't delete the quota config when the user set the quota to negative value. I think we should modify the "diskquota.show_fast_role_quota_view" as it considers all the configs in the table as valid.
I think it's better to delete the config when the user setting the quota to the negative value. 
And how about the UPDATE operation? If there is already a schema tablespace quota config in the table, and we insert another one with a different id but the same tablespace id and type, how does it work?

xuebinsu: 1. Agree. Will delete all negative configs.
2. Need to modify the UDF to implement UPDATE.

quotamodel.c
```diff
@@ -495,6 +656,7 @@ refresh_disk_quota_usage(bool is_init)
 	bool		pushed_active_snap = false;
 	bool		ret = true;
 
+	elog(WARNING, "refresh diskquota usage...");
```
xuebinsu, line 659:
```suggestion
	elog(LOG, "refresh diskquota usage...");
```

quotamodel.c
```diff
@@ -1074,115 +1106,130 @@ load_quotas(void)
...
 	 * config change.
 	 */
-	/* clear entries in quota limit map */
-	hash_seq_init(&iter, namespace_quota_limit_map);
-	while ((quota_entry = hash_seq_search(&iter)) != NULL)
+	clear_all_quota_maps();
+	const unsigned int NUM_ATTRIBUTES = 4;
+
+	/*
+	 * read quotas from diskquota.quota_config and target table
```
xuebinsu, line 1122: Do we still need to delete negative configs before reloading quotas?

fanfuxiaoran: When the user sets the quota value to be negative, the related record in the quota_config table will be deleted. Is there any chance for them to be negative?

PR #62, opened 2021-08-10
Support tablespace per segment ratio

fanfuxiaoran:
* support to set tablespace per segment
ratio to limit disk usage on segment.

SELECT diskquota.set_per_segment_quota('%tablespace_name%', ratio)

* update upgrade_test, support to test
old version extension scripts with new
version diskquota so.


quotamodel.c
```diff
@@ -148,14 +164,14 @@ static shmem_startup_hook_type prev_shmem_startup_hook = NULL;
 
 /* functions to maintain the quota maps */
 static void init_all_quota_maps(void);
-static void update_size_for_quota(int64 size, QuotaType type, Oid* keys);
-static void update_limit_for_quota(int64 limit, QuotaType type, Oid* keys);
+static void update_size_for_quota(int64 size, int64 maxsegsize, QuotaType type, Oid* keys);
+static void update_limit_for_quota(int64 limit, float4 segratio, QuotaType type, Oid* keys);
```
HaozhouWang, line 168: float4 is equal float?
It is better to use C standard type

quotamodel.c
```diff
@@ -1109,48 +1161,55 @@ do_load_quotas(void)
...
-	if (targetTableOid == InvalidOid)
+	if (strlen(extversion) < 3)
 	{
-		ret = SPI_execute("select targetoid, quotatype, quotalimitMB, 0 as tablespaceoid from diskquota.quota_config", true, 0);
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				errmsg("[diskquota] invalid diskquota extesion version: %s", extversion)));
 	}
-	else
+	if (memcmp(extversion, "1.0", 3) == 0)
```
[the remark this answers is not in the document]
fanfuxiaoran: Agreed, strcmp is better.

quotamodel.c
fanfuxiaoran: Yeah, in Postgres, float4 is equal to float.
```
typedef float float4;
```
And in DB, its type is real.

gp_activetable.c
```diff
@@ -800,11 +800,13 @@ pull_active_table_size_from_seg(HTAB *local_table_stats_map, char *active_oid_ar
 				/* receive table size info from the first segment */
 				entry->tableoid = tableOid;
 				entry->tablesize = tableSize;
+				entry->max_seg_tablesize = tableSize;
 			}
 			else
 			{
 				/* sum table size from all the segments */
 				entry->tablesize = entry->tablesize + tableSize;
+				entry->max_seg_tablesize = entry->max_seg_tablesize < tableSize ? tableSize : entry->max_seg_tablesize;
```
xuebinsu, line 809: Here `tableSize` is computed using `pg_total_relation_size()`, which might not be what we want.

fanfuxiaoran: Why? the total tableSize is the sum of tableSize from all segments computed by 
pg_total_relation_size() , and the max_seg_tablesize is the max size of tableSize from all segments computed by pg_total_relation_size.

"pg_total_relation_size" is executed only on segment. The only problem is that tableSize doesn't include the relation on master, but I think that's not a problem for us.

diskquota_utility.c
```diff
@@ -162,6 +170,46 @@ init_table_size_table(PG_FUNCTION_ARGS)
...
+		Oid         	oid;
+		int64       	sz;
+		int16		segid;
+
+		tup = SPI_tuptable->vals[i];
+		oid = SPI_getbinval(tup,tupdesc, 1, &isnull);
+		sz = SPI_getbinval(tup,tupdesc, 2, &isnull);
+		segid = SPI_getbinval(tup,tupdesc, 3, &isnull);
+
+		if (strcmp(extversion, "1.0") == 0)
```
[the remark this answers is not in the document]
fanfuxiaoran: Yeah,  now diskquota gets the ext major version as following:
```
return (int)strtol(extversion, (char **) NULL, 10);
```
I

xuebinsu (approved):
LGTM

PR #65, opened 2021-09-06
Fix index tbl

fanfuxiaoran:
Support index in a different tablespace with the original table

* Use pg_table_size instead of pg_total_realtion_size to get the disk usage.
   pg_table_size includes all forks, toast, and toast index, except the index.

* Add a function to get the list oids of the tables which the diskquota needs to care about in the database
firstly , get the tables which rekind is “r” or “m” and not system tables
```
 "select oid from pg_class"
                        " where oid >= %u and (relkind='r' or relkind='m')",
                        FirstNormalObjectId);
```
then fetch the indexes of those tables and add them to the list.
Use the list of oids in init_table_size_table and calculate_table_disk_usage functions

### DIFF
--- a/sql/test_index.sql
+++ b/sql/test_index.sql
@@ -13,18 +13,29 @@ SELECT diskquota.set_schema_tablespace_quota('indexschema1', 'indexspc','2 MB');
 SELECT pg_sleep(5);
 SELECT schema_name,tablespace_name,quota_in_mb,nspsize_tablespace_in_bytes FROM diskquota.show_fast_schema_tablespace_quota_view WHERE schema_name='indexschema1' and tablespace_name='indexspc';
 SELECT size, segid FROM diskquota.table_size , pg_class where tableid=oid and relname='test_index_a' and segid=-1;
--- create index for the table
+-- create index for the table, index in default tablespace
 CREATE INDEX a_index ON test_index_a(i);
 INSERT INTO test_index_a SELECT generate_series(1,10000);
 SELECT pg_sleep(5);
 -- expect insert success
 INSERT INTO test_index_a SELECT generate_series(1,100);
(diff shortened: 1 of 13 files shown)


sql/test_index.sql
```diff
@@ -0,0 +1,35 @@
...
+SET search_path TO indexschema1;
+
+CREATE TABLE test_index_a(i int) TABLESPACE indexspc;
+INSERT INTO test_index_a SELECT generate_series(1,20000);
+SELECT diskquota.set_schema_tablespace_quota('indexschema1', 'indexspc','2 MB');
+SELECT pg_sleep(5);
+SELECT schema_name,tablespace_name,quota_in_mb,nspsize_tablespace_in_bytes FROM diskquota.show_fast_schema_tablespace_quota_view WHERE schema_name='indexschema1' and tablespace_name='indexspc';
+SELECT size, segid FROM diskquota.table_size , pg_class where tableid=oid and relname='test_index_a' and segid=-1;
+-- create index for the table
+CREATE INDEX a_index ON test_index_a(i);
```
xuebinsu, line 17: Do we test the case where a query is rejected because only the index size exceeds some quota limit while the table size does NOT exceed any quota limit?

fanfuxiaoran: As the size of index is always small than the size of normal tables, so if the index and table are in the same tablespace, there is no this situation.
I can and a test case in which the index is in a different tablespace with the normal table, and set a quota for it .

xuebinsu: That's great. In fact, I doubt whether indexes are in the `rangeTable` when using the permission checking hook.

fanfuxiaoran: Good point！

fanfuxiaoran: You are correct,  it doesn't check the indexes quota.  When a tablespace only has indexes in it, even if the quota has exceeded, it can not block the query which inserts data in the index.
In the latest commit,  fetching the index list for each RangeTable, and checking the quota for the indexes.

xuebinsu (approved):
+1

PR #69, opened 2021-09-16
Add support for pausing and resuming the extension.

higuoxing:
This change adds two additional functions for user calling:

1. diskquota.pause_extension()
   After this function being called, diskquota keeps counting
   the disk usage but doesn't emit an error when the disk usage
  limit is exceeded.

2. diskquota.resume_extension()
   After this function being called, diskquota resumes to emit
   an error when the disk usage limit is exceeded.

### DIFF
--- a/quotamodel.c
+++ b/quotamodel.c
@@ -402,10 +402,11 @@ disk_quota_shmem_startup(void)
 	init_lwlocks();
 
 	/*
-	 * Three shared memory data. extension_ddl_message is used to handle
+	 * Four shared memory data. extension_ddl_message is used to handle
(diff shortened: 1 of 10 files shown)


quotamodel.c
```diff
@@ -1370,14 +1382,16 @@ quota_check_common(Oid reloid)
 		entry = hash_search(disk_quota_black_map,
 					&keyitem,
 					HASH_FIND, &found);
-		if (found)
+		if (found && !paused)
```
xuebinsu, line 1385: Maybe it is better to move this test outside of the loop and test it only once.

PR #71, opened 2021-09-28
Fix incorrect number of locks in diskquota.

higuoxing:
Currently, we use DiskQuotaLocksItemNumber to count the number of
locks in diskquota. In de8418b, we forget to update it and it's causing
memory issues. To avoid such issues in future, we use
sizeof(DiskQuotaLocks) / sizeof(void*) to calculate the correct number
of locks.

Co-authored-by: Hao Zhang <hzhang2@vmware.com>

### DIFF
--- a/diskquota.h
+++ b/diskquota.h
@@ -28,7 +28,6 @@ typedef enum
 	DISKQUOTA_READY_STATE
 }			DiskQuotaState;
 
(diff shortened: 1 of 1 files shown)


zhrt123 (commented):
OK

xuebinsu (approved):
LGTM

xuebinsu:
It's better to move the macro definition below the `typedef` since the macro now depends on the type.

beeender (approved):
:clap:

PR #75, opened 2021-10-14
Create database 'diskquota' before running regression tests.

higuoxing:
We have to create database 'diskquota' manually before running
regression tests. This change helps resolve it by creating the
database in the init stage.


xuebinsu (approved):
+1

PR #81, opened 2021-11-11
Fix stack full issue when flush to table_size

xuebinsu:


quotamodel.c
```diff
@@ -967,28 +967,18 @@ flush_to_table_size(void)
...
-			break;
-		case 2:
-			appendStringInfo(&delete_statement, "delete from diskquota.table_size where (tableid, segid) in ( ");
-			break;
-		default:
-			ereport(ERROR,
-					(errcode(ERRCODE_INTERNAL_ERROR),
-					 errmsg("[diskquota] unknown diskquota extension version: %d", extMajorVersion)));
-	}
+	optimizer = false;
```
higuoxing, line 978: Can we set `optimizer = false;` at the initialization stage of the diskquota bgworker, e.g., set it in `disk_quota_worker_main()`?

xuebinsu: Fixed by restoring `optimizer` and the end of this function.

Sasasu (approved):

higuoxing (approved):
+1

PR #82, opened 2021-11-16
Improve error handling in diskquota_fetch_table_stat().

higuoxing:
Historically, diskquota fetches tables' size by applying pg_table_size()
on a given oid list. It may invoke pg_table_size() several times in one
transaction. In order to tolerate exceptions, we wrap it with PG_TRY()
and consume any errors throwed by pg_table_size() using FlushErrorState().
It has the following 2 potential risks.

1. Deadlock: Considering we have a table 't1' (reloid=oid1) and it has a
   partition 't1_1_prt_11' (reloid=oid2).
   When diskquota_fetch_table_stat(1, '{oid2, oid1}'::oid[]) is called
   in session 'A', this UDF will invoke pg_table_size() on 't1_1_prt_11'
   and 't1' respectively. When pg_table_size('t1_1_prt_11') throws an
   exception, the AccessShareLock acquired from session 'A' will be held
   until the whole transaction ends. At the same time, when another
   session 'B' is performing 'ALTER TABLE t1', it will acquire the
   AccessExclusiveLock on 't1' and 't1_1_prt_11' respectively. As a result,
   the session 'A' is holding the AccessShareLock on 't1_1_prt_11' and is
   waiting for AccessShareLock on 't1'; the session 'B' is holding the
   AccessExclusiveLock on 't1' and is waiting for the AccessExclusiveLock
   on 't1_1_prt_11'.

2. Error message is missing. We should preserve the error message by calling
   CopyErrorData() before FlushErrorState() and elog(WARNING, error_messgae).

Co-Authored-by: Xuebin Su <sxuebin@vmware.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,5 +16,18 @@ REGRESS_OPTS = --schedule=diskquota_schedule_int --init-file=init_file
 else
 REGRESS_OPTS = --schedule=diskquota_schedule --init-file=init_file
 endif
+
+# FIXME: This check is hacky, since test_fetch_table_stat relies on the
+# gp_inject_fault extension, we detect if the extension is built with
+# greenplum by checking the output of the command 'pg_config --configure'.
+# In the future, if the diskquota is built with GPDB7, or we backport the
+# commit below to 6X_STABLE, we don't need this check.
+# https://github.com/greenplum-db/gpdb/commit/8b897b12f6cb13753985faacab8e4053bf797a8b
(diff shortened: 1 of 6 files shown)


xuebinsu (commented):
It will be great if we can have some test cases for this.

higuoxing:
I added one testcase by enabling 'gp_inject_fault' extension in gpdb. I'm still investigating how to disable faultinjector related tests in diskquota when the faultinjector is not enabled.

xuebinsu (approved):
+1

PR #84, opened 2021-11-21
Don't take relation's lock when fetching the relation's size.

higuoxing:
Currently, diskquota uses pg_table_size() to fetch relation's size. It
has the following drawbacks:

1. pg_table_size() requires taking the AccessShareLock of a relation and
   its indexes. When the relation is locked by AccessExclusive lock from
   another backend, the diskquota is blocked until the AccessShareLock
   is available. Besides, it also introduces potential deadlock to
   diskquota, since it can take multiple AccessShareLocks on different
   relations at the same time.

2. When we apply pg_table_size() on appendonly relations, it cannot give
   the actual file size when the relation is extending due to MVCC.

3. Fetching relations' size from segment servers and the master server
   shares the same logic, but we duplicate the codes here and there.

This patch resolves the mentioned issues by introducing a new helper
function 'collect_relation_size(HTAB *local_table)'. It takes a hash
table as argument and iterates over the pg_class table to acquire the
relfilenode for each relation. Then, it get the relation's size by
stat(2) the relfilenode directly and accumulate the size in local_table.

Co-Authored-By: Xuebin Su <sxuebin@vmware.com>
Co-Authored-By: Hao Zhang <hzhang2@vmware.com>

### DIFF
--- a/gp_activetable.c
+++ b/gp_activetable.c
@@ -13,12 +13,17 @@
  *
  * -------------------------------------------------------------------------
  */
+#include <sys/stat.h>
+
 #include "postgres.h"
 
 #include "access/htup_details.h"
 #include "access/xact.h"
 #include "catalog/indexing.h"
+#include "catalog/namespace.h"
 #include "catalog/pg_class.h"
+#include "catalog/pg_namespace.h"
+#include "catalog/pg_tablespace.h"
 #include "catalog/pg_type.h"
 #include "cdb/cdbbufferedappend.h"
 #include "cdb/cdbdisp_query.h"
@@ -39,6 +44,7 @@
 #include "utils/fmgroids.h"
 #include "utils/lsyscache.h"
 #include "utils/relfilenodemap.h"
+#include "utils/syscache.h"
 
 #include "gp_activetable.h"
 #include "diskquota.h"
@@ -71,6 +77,7 @@ static void pull_active_table_size_from_seg(HTAB *local_table_stats_map, char *a
 static StringInfoData convert_map_to_string(HTAB *active_list);
 static void load_table_size(HTAB *local_table_stats_map);
 static void report_active_table_helper(const RelFileNodeBackend *relFileNode);
+static Oid get_primary_oid(Oid reloid, Oid relnamespace, const char* relname);
 
 void		init_active_table_hook(void);
 void		init_shm_worker_active_tables(void);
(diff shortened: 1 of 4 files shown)


gp_activetable.c
```diff
@@ -937,3 +901,178 @@ pull_active_table_size_from_seg(HTAB *local_table_stats_map, char *active_oid_ar
...
+				{
+					/* TODO: Add tests for appendonly relations. */
+					if (strncmp(relname, "pg_aoseg_", 9))
+						key.reloid = atoi(&relname[10]);
+					else if (strncmp(relname, "pg_aocsseg_", 11))
+						key.reloid = atoi(&relname[12]);
+					else if (strncmp(relname, "pg_aoblkdir_", 12))
+						key.reloid = atoi(&relname[13]);
+					else if (strncmp(relname, "pg_aovisimap_", 13))
+						key.reloid = atoi(&relname[14]);
```
higuoxing, line 1040:
```suggestion
					if (strncmp(relname, "pg_aoseg_", 9) == 0)
						key.reloid = atoi(&relname[10]);
					else if (strncmp(relname, "pg_aocsseg_", 11) == 0)
						key.reloid = atoi(&relname[12]);
					else if (strncmp(relname, "pg_aoblkdir_", 12) == 0)
						key.reloid = atoi(&relname[13]);
					else if (strncmp(relname, "pg_aovisimap_", 13) == 0)
						key.reloid = atoi(&relname[14]);
```

Silly me!

gp_activetable.c
```diff
@@ -937,3 +910,184 @@ pull_active_table_size_from_seg(HTAB *local_table_stats_map, char *active_oid_ar
...
+			struct stat fst;
+
+			CHECK_FOR_INTERRUPTS();
+
+			if (segcount == 0)
+				snprintf(pathname, MAXPGPATH, "%s", relationpath);
+			else
+				snprintf(pathname, MAXPGPATH, "%s.%u", relationpath, segcount);
+
+			if (stat(pathname, &fst) < 0)
```
Sasasu, line 940: kindly remind: append-only tables with columnar storage do not have a sequential segment number.

```
create extension if not exists pgcrypto;
create table t(t1 text, t2 int) with (appendonly=true, orientation=column) distributed by (t2);
insert into t select gen_random_bytes(1024) from generate_series(1, 100000);
```
and check the name of the data file.

xuebinsu: We can use ao_foreach_extent_file() to enumerate all possible segment numbers. See https://github.com/greenplum-db/diskquota/blob/8b260f1925e26efbf7dba408d07433f4b71ffa6e/diskquota_utility.c#L1261

Sasasu: https://github.com/greenplum-db/diskquota/pull/93#discussion_r758933625

higuoxing:
Fixed in #100, close.

PR #85, opened 2021-11-22
Add test case for appendonly relations.

higuoxing:
Currently, tests for appendonly relations is missing. This patch adds
one test for that.

Co-Authored-by: Xuebin Su <sxuebin@vmware.com>


xuebinsu (approved):
+1

PR #92, opened 2021-11-29
Introduce isolation2 test framework to diskquota.

higuoxing:
This patch introduces isolation2 test framework to diskquota. We are
able to write concurrent test cases for diskquota after this patch.


xuebinsu (approved):
+1

PR #96, opened 2021-12-01
fix bug: acquire lock when traversing relation_cache

zhrt123:
We should acquire relation_cache_lock when traversing relation_cache in remove_committed_relation_from_cache()


xuebinsu (approved):
+1

PR #97, opened 2021-12-01
Fix test bug: drop table after testing relation_size()

zhrt123:


higuoxing:
It looks that you haven't fix the .out file?

PR #99, opened 2021-12-02
Fix missing update_diskquota_db_list()

xuebinsu:
Known issue: 
- There might be garbage active oids in databases with Diskquota not enabled and cannot be deleted by any bgworker.


tests/regress/sql/test_udf_update_db.sql
```diff
@@ -0,0 +1,6 @@
+-- Test if this UDF is created automatically after restarting GPDB.
+-- Note: This test must be placed before all quota tests.
```
higuoxing, line 2: I think we can place this test anywhere by checking the description of the function.

```
postgres=# \df diskquota.update_diskquota_db_list
```

xuebinsu: No, we can't, because connection will be terminated after `gpstop` and the `\df` must be placed in another `.sql` file.

higuoxing: https://github.com/greenplum-db/gpdb/blob/master/contrib/sslinfo/sql/sslinfo.sql#L7 Haven't tested it yet, but I think it's worth a try.

beeender:
I am wondering why would `monitoring_dbid_cache` ever be useful?
- If the extension has not been created on a database, then the hook `init_active_table_hook` has never been called
- If the extension has been created, but then dropped, we should just de-register the `init_active_table_hook` 
It seems we never need a `monitoring_dbid_cache` at all?

beeender:
>  Actually, if we do the de-regestering in the _PG_fini(), then everything should just work?

Hmmmm. no, it won't work since the so is in `shared_preload_libraries`.

xuebinsu:
Closed in favor of PR #138 .

PR #101, opened 2021-12-06
Fix incorrect JOIN condition in test_blackmap.sql

higuoxing:
This PR fixes incorrect JOIN condition in test_blackmap.sql. The
relfilenodes are not always different across segments. Hence, we should
add an additional JOIN condition to the test case or it will produce
unstable results.


xuebinsu (approved):
+1

zhrt123 (approved):
+1

PR #102, opened 2021-12-07
Add support for adding uncommitted relations to blackmap.

higuoxing:
This patch adds support for adding uncommitted relations to blackmap on
segment servers. Most of the codes share similar logic with adding
committed relations to blackmap. Test will be added in the next
following commits.


xuebinsu (dismissed):
+1

xuebinsu (approved):
+1

PR #106, opened 2021-12-13
Ignore distribution notice in regression tests.

higuoxing:
This PR is trying to fix CI building failure by ignoring distribution
notice in CTAS statements.


xuebinsu (approved):
+1

PR #112, opened 2021-12-20
Fix bug: table's relation_cache_entry is not be removed after vacuum full

zhrt123:
The modified table's relation_cache_entry is not be removed after vacuum full.
Use RelidByRelfilenode() to check whether the table is committed, and remove its relation_cache_entry.

### DIFF
--- a/relation_cache.c
+++ b/relation_cache.c
@@ -289,9 +289,14 @@ remove_committed_relation_from_cache(void)
 	hash_seq_init(&iter, local_relation_cache);
 	while ((local_entry = hash_seq_search(&iter)) != NULL)
 	{
-		if (SearchSysCacheExists1(RELOID, local_entry->relid))
+		/* The committed table's oid can be fetched by RelidByRelfilenode().
+		 * If the table's relfilenode is modified and its relation_cache_entry
+		 * remains in relation_cache, the outdated relation_cache_entry should 
+		 * be removed.
+		 */
+		if (OidIsValid(RelidByRelfilenode(local_entry->rnode.node.spcNode, local_entry->rnode.node.relNode)))
 		{
-			remove_cache_entry(local_entry->relid, InvalidOid);
+			remove_cache_entry(InvalidOid, local_entry->rnode.node.relNode);
 		}
 	}
 	hash_destroy(local_relation_cache);
--- a/relation_cache.c
+++ b/relation_cache.c
@@ -289,7 +289,8 @@ remove_committed_relation_from_cache(void)
 	hash_seq_init(&iter, local_relation_cache);
 	while ((local_entry = hash_seq_search(&iter)) != NULL)
 	{
-		/* The committed table's oid can be fetched by RelidByRelfilenode().
+		/*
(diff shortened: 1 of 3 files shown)


relation_cache.c
```diff
@@ -289,9 +289,14 @@ remove_committed_relation_from_cache(void)
...
-		if (SearchSysCacheExists1(RELOID, local_entry->relid))
+		/* The committed table's oid can be fetched by RelidByRelfilenode().
+		 * If the table's relfilenode is modified and its relation_cache_entry
+		 * remains in relation_cache, the outdated relation_cache_entry should 
+		 * be removed.
+		 */
+		if (OidIsValid(RelidByRelfilenode(local_entry->rnode.node.spcNode, local_entry->rnode.node.relNode)))
 		{
-			remove_cache_entry(local_entry->relid, InvalidOid);
+			remove_cache_entry(InvalidOid, local_entry->rnode.node.relNode);
```
xuebinsu, line 299: Do we still need to `remove_cache_entry()` by relid?

relation_cache.c
```diff
@@ -289,9 +289,14 @@ remove_committed_relation_from_cache(void)
 	hash_seq_init(&iter, local_relation_cache);
 	while ((local_entry = hash_seq_search(&iter)) != NULL)
 	{
-		if (SearchSysCacheExists1(RELOID, local_entry->relid))
+		/* The committed table's oid can be fetched by RelidByRelfilenode().
```
higuoxing, line 292: Nit: comment style.
```suggestion
		/*
		 * The committed table's oid can be fetched by RelidByRelfilenode().
```

tests/regress/expected/test_vacuum.out
```diff
@@ -34,6 +34,19 @@ SELECT pg_sleep(20);
  
 (1 row)
 
+SELECT tableid::regclass, size, segid from diskquota.table_size WHERE tableid::regclass::name not like '%.%' ORDER BY size, segid DESC;
```
higuoxing, line 37: Nit: Keep the SQL statements consistent across this file.
```suggestion
SELECT tableid::regclass, size, segid from diskquota.table_size WHERE tableid::regclass::name NOT LIKE '%.%' ORDER BY size, segid DESC;
```

higuoxing (dismissed):
This change looks good to me. Please land it once @xuebinsu is happy with it.

relation_cache.c
zhrt123: In get_relfilenode_by_relid(), we should use relid to remove cached entry.

xuebinsu: Why do we still need to remove cache entry in `get_relfilenode_by_relid()`

zhrt123: We use relid to fetch relfilenode. If the relfilenode in relation_cache is wrong, we cannot use the correct relfilenode to remove the wrong cached entry. But we can use the relid to remove this entry.

xuebinsu: OK, it makes sense.

xuebinsu (approved):
+1

PR #117, opened 2022-01-10
ci: create rhel8 release build.

Sasasu:
add rhel8 support

### DIFF
--- a/concourse/pipeline/pipeline.yml
+++ b/concourse/pipeline/pipeline.yml
@@ -94,7 +94,7 @@ resources:
 - name: diskquota_src
   type: git
   source:
-    branch: release/1.X
+    branch: gpdb
     uri: https://github.com/greenplum-db/diskquota.git
 
 # gpdb binary on gcs is located as different folder for different version
--- a/concourse/pipeline/pipeline.yml
+++ b/concourse/pipeline/pipeline.yml
@@ -13,6 +13,7 @@ groups:
   jobs:
     - diskquota_centos6_build_test
     - diskquota_centos7_build_test
+    - diskquota_rhel8_build_test
     - diskquota_ubuntu18_build_test
 - name: GPDB6_UPGRADE
   jobs:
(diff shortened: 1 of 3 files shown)


concourse/scripts/build_diskquota.sh
```diff
@@ -41,6 +41,13 @@ function pkg() {
         share/postgresql/extension/diskquota--2.0--1.0.sql \
         install_gpdb_component
       ;;
+    rhel8)
+        tar -czf $TOP_DIR/diskquota_artifacts/diskquota-${DISKQUOTA_VERSION}-rhel8_x86_64.tar.gz \
+        lib/postgresql/diskquota.so \
+        share/postgresql/extension/diskquota.control \
+        share/postgresql/extension/diskquota--1.0.sql \
```
higuoxing, line 48: Two comments from myself.
1. Looks that `diskquota--1.0--2.0.sql`, `diskquota--2.0--1.0.sql` and `diskquota--2.0.sql` are missing?
2. I think the script can be simplified (we can do it in a follow-up change).
    ```
    case "$DISKQUOTA_OS" in
    rhel6 | rhel7 | ubuntu18.04)
      ## Package the diskquota.
      ;;
    ```

higuoxing (approved):
LGTM. Thanks!

concourse/pipeline/pipeline.yml
```diff
@@ -77,7 +94,7 @@ resources:
 - name: diskquota_src
   type: git
   source:
-    branch: gpdb
+    branch: rel/for-gpdb-6x
```
beeender, line 97: don't forget change this before merging

Sasasu: this change is on purpose.

`gpdb` branch current is version `2.X`, can not use that branch to release new 1.X version.
so xuebin and me decide to remove `gpdb` branch, use `rel/for-gpdb-6x` and `rel/for-gpdb-7x`.

beeender: `rel/for-gpdb-6x`
I am confused, is this branch for release diskquota 1.x? Why not just call it `release/1.x`?

Sasasu: how about `release/6X/1.x`
cc @xuebinsu

beeender: I feel like we should use the same source code to release 6x/7x. but that might be difficult for diskquota? We can have a discussion about that.

higuoxing: No, I tried to compile diskquota with gpdb 7X several months ago. There're some compiling errors due to inconsistent APIs between 6X and 7X and I didn't dig into it too much.

beeender: I think we can manage the compiles by macros. Using separated source code for 6x/7x will be a headache for maintainance.

higuoxing: Sounds reasonable. So we're going to have two branches (maybe `dev/1.X` and `dev/2.X`?) for development and two branches for releasing (e.g. `release/2.X` and `release/1.X`)?

beeender: We are going to have only one development branch. `1.x` won't have any minor releases, only patch releases if needed.
All new features will go to the dev branch as `2.x.x`.

beeender (approved):

PR #118, opened 2022-01-18
Wait until deleted tuples are dead before VACUUM

xuebinsu:
Consider a user session that does a DELETE followed by a VACUUM FULL to
reclaim the disk space. If, at the same time, the bgworker loads config
by doing a SELECT, and the SELECT begins before the DELETE ends, while ends
after the VACUUM FULL begins:

bgw: ---------[ SELECT ]----------->
usr: ---[ DELETE ]-[ VACUUM FULL ]-->

then the tuples deleted will be marked as RECENTLY_DEAD instead of DEAD.
As a result, the deleted tuples cannot be removed by VACUUM FULL.

The fix lets the user session wait for the bgworker to finish the current
SELECT before starting VACUUM FULL .

### DIFF
--- a/tests/regress/expected/test_partition.out
+++ b/tests/regress/expected/test_partition.out
@@ -45,6 +45,12 @@ ERROR:  schema's disk space quota exceeded with name:s8
 INSERT INTO measurement SELECT 1, '2006-03-03' ,1,1;
 ERROR:  schema's disk space quota exceeded with name:s8
 DELETE FROM measurement WHERE logdate='2006-03-02';
+SELECT diskquota.wait_for_worker_new_epoch();
+ wait_for_worker_new_epoch 
+---------------------------
+ t
+(1 row)
+
 VACUUM FULL measurement;
 SELECT diskquota.wait_for_worker_new_epoch();
  wait_for_worker_new_epoch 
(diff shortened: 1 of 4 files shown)


higuoxing (approved):
LGTM. ☺️

tests/regress/expected/test_partition.out
```diff
@@ -45,6 +45,12 @@ ERROR:  schema's disk space quota exceeded with name:s8
 INSERT INTO measurement SELECT 1, '2006-03-03' ,1,1;
 ERROR:  schema's disk space quota exceeded with name:s8
 DELETE FROM measurement WHERE logdate='2006-03-02';
+SELECT diskquota.wait_for_worker_new_epoch();
```
xuebinsu, line 48:
```suggestion
SELECT diskquota.wait_for_worker_new_epoch();
```

xuebinsu:
Re-trigger CI

beeender (approved):

fanfuxiaoran:
Where does the diskquota bgw run a query on the user table?

xuebinsu:
@fanfuxiaoran 

> Where does the diskquota bgw run a query on the user table?

The bgworker only accesses tables under the `diskquota` schema. But it will still affect `VACUUM FULL` because the tuple visibility only depends on the distributed transaction id, regardless of which table is being accessed.

PR #119, opened 2022-01-20
Refactor pause() to skip refreshing quota

xuebinsu:
Currently, diskquota.pause() only takes effect on quota checking.
Bgworkers still go over the loop to refreshing quota even if diskquota
is paused. This wastes computation resources and can cause flaky issues.

This fix makes bgworkers skip refreshing quota when the user pauses
diskquota entirely to avoid those issues. Table sizes can be updated
correctly after resume.

### DIFF
--- a/diskquota.c
+++ b/diskquota.c
@@ -79,6 +79,15 @@ static int	num_db = 0;
  */
 bool *diskquota_paused = NULL;
 
+bool
+diskquota_is_paused()
+{
+	LWLockAcquire(diskquota_locks.paused_lock, LW_SHARED);
+	bool paused = *diskquota_paused;
+	LWLockRelease(diskquota_locks.paused_lock);
+	return paused;
+}
+
 /* functions of disk quota*/
 void		_PG_init(void);
 void		_PG_fini(void);
@@ -365,7 +374,8 @@ disk_quota_worker_main(Datum main_arg)
 		}
 
 		/* Do the work */
-		refresh_disk_quota_model(false);
+		if (!diskquota_is_paused())
+			refresh_disk_quota_model(false);
 		worker_increase_epoch(MyDatabaseId);
 	}
 
--- a/tests/regress/diskquota_schedule
+++ b/tests/regress/diskquota_schedule
@@ -4,6 +4,7 @@ test: test_worker_epoch
 test: test_relation_size
 test: test_relation_cache
 test: test_uncommitted_table_size
+test: test_pause_and_resume
 # disable this tese due to GPDB behavior change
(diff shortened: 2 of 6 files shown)


diskquota.c
```diff
@@ -79,6 +79,16 @@ static int	num_db = 0;
  */
 bool *diskquota_paused = NULL;
 
+static inline bool
+is_worker_paused()
+{
+
```
higuoxing, line 85: Nit: remove this blank line.
```suggestion
```

higuoxing (approved):
LG!

tests/regress/diskquota_schedule
```diff
@@ -4,6 +4,7 @@ test: test_worker_epoch
 test: test_relation_size
 test: test_relation_cache
 test: test_uncommitted_table_size
+test: test_pause_and_resume
```
beeender, line 7: any special reason we move this case here?

diskquota.c
```diff
@@ -79,6 +79,16 @@ static int	num_db = 0;
...
+	LWLockAcquire(diskquota_locks.paused_lock, LW_SHARED);
```
beeender, line 86: We can simply remove the `pause_lock` by using pg's atomic operations.
https://github.com/greenplum-db/gpdb/blob/master/src/include/port/atomics.h

It is not related to this PR. If needed, we can do in another PR.

tests/regress/diskquota_schedule
xuebinsu: Because `pause()` and `resume()` are fundamental functions and need to be used in other tests. 

For example, we need to `SELECT diskquota.pause();` before `DROP EXTENSION diskquota;` to avoid possible deadlocks in `test_extension`.

diskquota.c
xuebinsu: Good idea. In fact, we have some other shared-memory variables that needs to be refactored that way such as `diskquota_hardlimit`. Let's leave it to the next PR.

tests/regress/diskquota_schedule
higuoxing: I think whether `pause()` or not shouldn't affect dropping the extension. Not sure if we can fix it in https://github.com/greenplum-db/diskquota/blob/d04e5ef1bebe11534bed763786e49ea4bb4d686c/diskquota_utility.c#L449 by pausing the extension automatically before dropping it.

xuebinsu: @higuoxing Good idea. will check the hook later.

beeender (approved):
:+1:

PR #121, opened 2022-01-24
Pause before DROP EXTENSION

xuebinsu:
Currently, deadlock can occur when

- A user session is doing DROP EXTENSION, and
- A bgworker is loading quota configs using SPI.

This patch fixes the issue by pausing diskquota before DROP 
EXTENSION so that the bgworker will not load config anymore.

Note that this cannot be done using object_access_hook() because
the extension object is dropped AFTER dropping all tables that belong
to it.

### DIFF
--- a/tests/isolation2/sql/cleanup.sql
+++ b/tests/isolation2/sql/cleanup.sql
@@ -1,3 +1,5 @@
 -- start_ignore
+SELECT diskquota.pause();
+SELECT diskquota.wait_for_worker_new_epoch();
 DROP EXTENSION diskquota;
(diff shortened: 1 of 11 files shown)


beeender:
Does this one need #122 to be merged first?

xuebinsu:
> Does this one need #122 to be merged first?

@beeender Yes.

Sasasu (approved):

Sasasu (approved):

Sasasu (approved):

beeender (approved):

PR #122, opened 2022-01-24
make diskquota_is_paused() database level

Sasasu:


diskquota.c
```diff
@@ -73,18 +73,33 @@ HTAB *disk_quota_worker_map = NULL;
...
 
 /*
- * diskquota_paused is a flag used to pause the extension (when the flag is
- * enabled, the extension keeps counting the disk usage but doesn't emit an
- * error when the disk usage limit is exceeded).
+ * if diskquota_paused is true.
+ * adding new database will use diskquota_paused=true as default
  */
-bool *diskquota_paused = NULL;
+pg_atomic_uint32 *diskquota_paused = NULL;
```
xuebinsu, line 79: Why not reomve this?

Sasasu: consider this:

1. user call `diskquota.pause()`
2. user creates diskquota in a new database.
3. the new database quota should be paused, this is the current logic.

if we remove `diskquota_paused`. consider this:
1. user does `vaccum full` in all diskquota enabled database. it means all `is_paused == true`
2. user creates diskquota in a new database.
3. the new database quota should be enabled as default. 

but we don't have enough conditions to detect this state.

beeender: can we have something like:
```
pg_atomic_uint32 **diskquota_paused = NULL;
void init() {
lock();
hash_entry = (DiskQuotaWorkerEntry*) hash_search(disk_quota_worker_map,
														(void*)&MyDatabaseId,
														HASH_FIND,
														&found);
unlock();
diskquota_paused = &hash_entry->is_paused;
}

```

Then we ever need to acquire lock before reading it.

Otherwise, the `hash_entry->is_paused` doesn't have to be atomic since it is always accessed with locks.

Sasasu: > can we have something like.
we can not, the hash map may extend and re-order.

> hash_entry->is_paused doesn't have to be atomic
👌

xuebinsu (approved):
+1

diskquota.c
```diff
@@ -72,19 +72,25 @@ ExtensionDDLMessage *extension_ddl_message = NULL;
...
+	LWLockAcquire(diskquota_locks.worker_map_lock, LW_SHARED);
+	{
+		DiskQuotaWorkerEntry	*hash_entry;
+		bool					found;
+
+		hash_entry = (DiskQuotaWorkerEntry*) hash_search(disk_quota_worker_map,
+														(void*)&MyDatabaseId,
+														HASH_FIND,
+														&found);
+		paused = found ? hash_entry->is_paused : false;
```
beeender, line 90: if the entry is not found, should it still do refresh?
I feel when the entry is found, the `paused` should be `true`?

diskquota_utility.c
```diff
@@ -372,12 +398,36 @@ diskquota_resume(PG_FUNCTION_ARGS)
...
+													(void*)&dbid,
+													HASH_FIND,
+													&found);
+		if (found) {
+			hentry->is_paused = false;
+		}
+
+		// remove the element since we do not need any more
+		// ref diskquota_pause()
+		if (found && hentry->handle == NULL) {
```
beeender, line 422: when will the `hentry->handle` be `NULL`? 
The only case I saw is that the entry was `memset` to `0` in the pause, then it was directly resumed before worker start. Right?
So it is used to solve the problem:
1) worker not started
2) pause
3) worker start.

Right? But in the `start_worker_by_dboid(Oid dbid)`

```
	/* put the worker handle into the worker map */
	workerentry = (DiskQuotaWorkerEntry *) hash_search(disk_quota_worker_map,
													   (void *) &dbid,
													   HASH_ENTER, &found);
	if (!found)
	{
		workerentry->handle = handle;
		workerentry->pid = pid;
		workerentry->epoch = 0;
	}

```
In this case, the `found` will be true and those won't be set. Correct?

Shall we make this simpler? Only creating the entry in one place, and report error when `pause`/`resume` called if the bgworker is not started?

diskquota.c
Sasasu: > if the entry is not found, should it still do refresh?

this is not related to refreshing.

there have 3 types of processes.
- launcher on the coordinator. | worker booted, nothing to do in this PR scope.
- worker on the coordinator. | worker booted, will do the refreshing.
- gang process on a segment. | worker not boot, and will check `diskquota_is_paused()`

this check will occur in type 2 and type 3.

- for worker on the coordinator: check if `diskquota_is_paused` stops refreshing.
- for gang process on a segment: check if `diskquota_is_paused` stop emits error.

type 2 is easy to understand, `found` will always be `true`

for type 3, the `disk_quota_worker_map` is empty if nobody called `diskquota.pause()`.
so we need the default value here. to let the code path go into `black map check`, we need `pause = false` here.

diskquota_utility.c
Sasasu: > I feel when the entry is found, the paused should be true?

worker process always exists on coordinator.

> when will the hentry->handle be NULL?

in gang process on a segment. the handle is empty.

-------

we are saving two data (`the worker process for database`  and `database is pause or not`) in one data struct, this is hard to understand :(

Sasasu:
I just `push -f` to squash the typo fix commit.

diskquota.c
beeender: Can we use `GpRoleValue Gp_role` for checking here? To make the logic easier to be understood and avoid creating this map on QE.
```
if (Gp_role == GP_ROLE_EXECUTE) {
    return true;
}
```

diskquota_utility.c
xuebinsu: What is the value of `is_paused` after this sequence of operations?
```sql
SELECT diskquota.pause();
DROP EXTENSION diskquota;
CREATE EXTENSION diskquota;
```
Is `is_paused` `false`?

beeender (approved):

xuebinsu (approved):

PR #123, opened 2022-01-25
Fix flaky test case in isolation2/test_relation_size.sql

higuoxing:
This PR is trying to fix a race condition in `isolation2/test_relation_size`. It's observed in our CI https://extensions.ci.gpdb.pivotal.io/teams/main/pipelines/diskquota-pr-gpdb-6x/jobs/test_diskquota_pr/builds/30

-- edited by beeender start--
The pipeline has been renamed:
https://extensions.ci.gpdb.pivotal.io/teams/main/pipelines/DEPRECATED:diskquota-pr-gpdb-6x/jobs/test_diskquota_pr/builds/30
-- edited by beeender end--

### DIFF
--- a/tests/isolation2/expected/test_relation_size.out
+++ b/tests/isolation2/expected/test_relation_size.out
@@ -17,29 +17,37 @@ SELECT diskquota.relation_size('t_dropped');
 (1 row)
 
 -- Inject 'suspension' to servers.
-SELECT gp_inject_fault_infinite('diskquota_before_stat_relfilenode', 'suspend', dbid) FROM gp_segment_configuration WHERE role='p';
+SELECT gp_inject_fault_infinite('diskquota_before_stat_relfilenode', 'suspend', dbid) FROM gp_segment_configuration WHERE role='p' AND content>=0;
  gp_inject_fault_infinite 
 --------------------------
  Success:                 
  Success:                 
  Success:                 
- Success:                 
-(4 rows)
+(3 rows)
 
 -- Session 1 will hang before applying stat(2) to the physical file.
 1&: SELECT diskquota.relation_size('t_dropped');  <waiting ...>
+-- Wait until the fault is triggered to avoid the following race condition:
+-- The 't_dropped' table is dropped before evaluating "SELECT diskquota.relation_size('t_dropped')"
+-- and the query will fail with 'ERROR: relation "t_dropped" does not exist'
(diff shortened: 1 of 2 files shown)


higuoxing:
Seems that CI isn't happy with this change. I'm going to fix it.

xuebinsu:
Could you please explain why this case is "flaky" rather than consistent?

higuoxing:
> Could you please explain why this case is "flaky" rather than consistent?

```sql
-- Inject 'suspend' fault into diskquota.relation_size().
-- ...

1&: SELECT diskquota.relation_size('t_dropped');
-- Drop the table.
DROP TABLE t_dropped;

-- Remove the 'suspend' fault.
-- ...

1<: -- Wait session 1 return.
```

In the above test, we want to make sure that when the actual relfilenode of relation `t_dropped` is removed before `stat(2)` the physical file, the `diskquota.relation_size()` will return 0 as the file size.

However, we only know that `1&: SELECT diskquota.relation_size('t_dropped'::regclass)` is executed before `DROP TABLE` and cannot guarantee that `'t_dropped'` is evaluated to `regclass` before `'t_dropped'` is dropped. As a result, the query in session 1 may fail with `'ERROR: relation "t_dropped" does not exist'`

```
   ----------------------------------------------------------------------------------------------> Time
s1: | Parse statement |                          ...        | Evaluate 't_dropped' to regclass |
s2:        | Parse statement | ... | Drop 't_dropped' |     ^ 't_dropped' doesn't exist!
```

The solution is that, before executing `DROP TABLE t_dropped`, use `gp_wait_until_triggered_fault()` to block the process until the fault is triggered. However, the CI isn't happy with this change, I'm going to fix it.

xuebinsu:
Got it! So we need to ensure that `t_dropped` is dropped before `stat(2)` but after getting its `relfilenode`.

Please retrigger CI by e.g., pushing an empty commit.

xuebinsu:
Status: ready for review

beeender:
@higuoxing can you please push an empty to trigger the pipeline?

higuoxing:
Looks that the pipeline is disabled for PR from external repo?

beeender:
> Looks that the pipeline is disabled for PR from external repo?

#156 added a doc there.

beeender (approved):
:clap:

xuebinsu (approved):

PR #124, opened 2022-01-25
send a prompt if launcher crashed

Sasasu:
find a possible long time wait:

1. launcher deaded or in dead lock
2. do `create extension`
3. the default timeout is 120s, which is very long.

this PR adds a warning message after 3.6s waiting.

also this PR remove a lock `diskquota_locks.hardlimit_lock` and clear up some `include`

### DIFF
--- a/diskquota_utility.c
+++ b/diskquota_utility.c
@@ -236,7 +236,7 @@ generate_insert_table_size_sql(StringInfoData *insert_buf, int extMajorVersion)
 Datum
 diskquota_start_worker(PG_FUNCTION_ARGS)
 {
-	int			rc;
+	int rc, launcher_pid;
 
 	/*
 	 * Lock on extension_ddl_lock to avoid multiple backend create diskquota
@@ -248,8 +248,9 @@ diskquota_start_worker(PG_FUNCTION_ARGS)
 	extension_ddl_message->cmd = CMD_CREATE_EXTENSION;
 	extension_ddl_message->result = ERR_PENDING;
 	extension_ddl_message->dbid = MyDatabaseId;
+	launcher_pid = extension_ddl_message->launcher_pid;
 	/* setup sig handler to diskquota launcher process */
-	rc = kill(extension_ddl_message->launcher_pid, SIGUSR1);
+	rc = kill(launcher_pid, SIGUSR1);
 	LWLockRelease(diskquota_locks.extension_ddl_message_lock);
 	if (rc == 0)
 	{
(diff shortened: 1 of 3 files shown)


xuebinsu:
How can this "long-time wait" happen?

diskquota_utility.c
```diff
@@ -275,6 +265,14 @@ diskquota_start_worker(PG_FUNCTION_ARGS)
 			if (rc & WL_POSTMASTER_DEATH)
 				break;
 			ResetLatch(&MyProc->procLatch);
+
+			// prompt 10 times, interval 12s
+			if (count <= WAIT_TIME_COUNT_WARNING && (count % (WAIT_TIME_COUNT_WARNING/10) == 0)) {
```
beeender, line 270: We probably can use a better way to check the launcher's death caused hang issue.
`rc = kill(launcher_pid, 0)` might be helpful here. And error out the loop.

beeender (commented):
Code deletion looks good :)

diskquota_utility.c
```diff
@@ -264,6 +264,11 @@ diskquota_start_worker(PG_FUNCTION_ARGS)
 			if (rc & WL_POSTMASTER_DEATH)
 				break;
 			ResetLatch(&MyProc->procLatch);
+
+			ereportif(ERROR,
+					kill(extension_ddl_message->launcher_pid, 0) == -1 && errno == ESRCH, // do existence check
```
beeender, line 269: will `!= 0` be safer than `== -1`?

beeender:
```
       For a process to have permission to send a signal, it must either be privileged (under Linux: have the CAP_KILL capability in the user namespace of the target process), or
       the  real or effective user ID of the sending process must equal the real or saved set-user-ID of the target process.  In the case of SIGCONT, it suffices when the sending
       and receiving processes belong to the same session.  (Historically, the rules were different; see NOTES.)

RETURN VALUE
       On success (at least one signal was sent), zero is returned.  On error, -1 is returned, and errno is set to indicate the error.

ERRORS
       EINVAL An invalid signal was specified.

       EPERM  The calling process does not have permission to send the signal to any of the target processes.

       ESRCH  The target process or process group does not exist.  Note that an existing process might be a zombie, a process that has terminated execution, but has not yet  been
              wait(2)ed for.
```

OK, this looks good

diskquota_utility.c
```diff
@@ -264,6 +264,11 @@ diskquota_start_worker(PG_FUNCTION_ARGS)
 			if (rc & WL_POSTMASTER_DEATH)
 				break;
 			ResetLatch(&MyProc->procLatch);
+
```
beeender, line 267: uh...
`extension_ddl_message` was protected in 245-253.
use a local var to save the pid in the lock area, and use the local var here. To avoid the concurrency problem.

diskquota_utility.c
```diff
@@ -522,6 +527,11 @@ dq_object_access_hook(ObjectAccessType access, Oid classId,
...
+			ereportif(ERROR,
```
beeender, line 531: ditto

beeender (approved):

xuebinsu (dismissed):
+1. We might need to re-consider whether it is necessary for the launcher to be long-lived. It seems that it is enough to create a launcher in `diskquota_start_worker()`.

beeender (approved):

xuebinsu (approved):

Sasasu:
CI failure due to a known flaky test case #121 will fix it.
let's ignore CI failure :)

PR #126, opened 2022-01-26
allow to re-create extension

Sasasu:
same as title.


xuebinsu (approved):
+1

xuebinsu:
I think a more convenient way to test this is to do `SELECT diskquota.diskquota_start_worker();` twice. (Need to remove `DROP FUNCTION` first.)

beeender (approved):
:clap:

PR #127, opened 2022-01-26
add diskquota.status() to show if hardlimit or softlimit enabled or not

Sasasu:
```
select * from diskquota.status();
    name    | status  
------------+---------
 soft limit | enabled
 hard limit | enabled
(2 rows)
```

### DIFF
--- a/diskquota.c
+++ b/diskquota.c
@@ -19,6 +19,7 @@
 
 #include "postgres.h"
 
+#include "funcapi.h"
 #include "access/xact.h"
 #include "cdb/cdbvars.h"
 #include "commands/dbcommands.h"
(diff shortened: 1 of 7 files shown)


tests/regress/expected/test_show_status.out
```diff
@@ -0,0 +1,51 @@
+select * from diskquota.status();
+    name    |  status   
+------------+-----------
+ soft limit | enabled
+ hard limit | disabled
+(2 rows)
```
Sasasu, line 6: hard limit is disabled by default, this is correct?
or only test environment is disabled by default?

higuoxing: Yes, it's disabled by default.

BTW, I think `diskquota.status()` might not be a good name, since the `status` is too general for the result. Maybe call it `diskquota.show_limit_types()` or `diskquota.show_limit_status()`?

xuebinsu:
I think it would be simpler to do
```sql
CREATE OR REPLACE VIEW diskquota.status AS
SELECT diskquota.is_paused(), diskquota.hard_limit();
```
Soft limit cannot be disabled, so I don't think it is necessary to include that.

Sasasu:
> Soft limit cannot be disabled

in fact, can be disabled by modifying `shared_preload_libraries`

xuebinsu:
> in fact, can be disabled by modifying shared_preload_libraries

Can the user check that using this UDF?

beeender:
What was the conclusion of our discussion? One more row like :

```
    name    |  status   
------------+-----------
 soft limit | enabled
 hard limit | disabled
 paused | no
```

Sasasu:
push -f to resolve conflict, not code change.

diskquota.c
```diff
@@ -1079,3 +1080,84 @@ show_worker_epoch(PG_FUNCTION_ARGS)
...
+		int index;
+	} Context;
+
+	typedef struct FeatureStatus {
+		const char* name;
+		const char* (*status)(void);
+	} FeatureStatus;
+
+	static const FeatureStatus fs[] = {
+		{.name = "soft limit", .status = diskquota_status_check_soft_limit},
```
xuebinsu, line 1124: I think `"soft limit"` here should be `"paused"` since it does not actually check `shared_preload_libraries`.

higuoxing:
Looks good! Please wait until Xuebin is happy with this change :-)

beeender (approved):

xuebinsu (approved):

PR #130, opened 2022-01-30
Add timeout when waiting for workers

xuebinsu:
Each time the state of Diskquota is changed, we need to wait for the
change to take effect using diskquota.wait_for_worker_new_epoch().
However, when the bgworker is not alive, such wait can last forever.

This patch fixes the issue by adding a timeout GUC so that wait() will
throw an NOTICE if it times out, making it more user-friendly.

To fix a race condition when CREATE EXTENSION, the user needs to 
SELECT wait_for_worker_new_epoch() manually before writing data.
This is to wait until the current database is added to the monitored
db cache so that active tables in the current database can be 
recorded.

This patch also fix test script for activating standby and rename
some of the cases to make them more clear.


diskquota.c
```diff
@@ -1079,3 +1098,48 @@ show_worker_epoch(PG_FUNCTION_ARGS)
...
+			errmsg("[diskquota] timeout when waiting for worker"),
+			errhint("please check if the bgworker is still alive.")));
+		return true;
+	}
+	return false;
+}
+
+PG_FUNCTION_INFO_V1(wait_for_worker_new_epoch);
+Datum
+wait_for_worker_new_epoch(PG_FUNCTION_ARGS)
```
Sasasu, line 1120: why not just use

```
PG_FUNCTION_INFO_V1(wait_for_worker_new_epoch);
Datum wait_for_worker_new_epoch(PG_FUNCTION_ARGS)
{
    TimestampTz start_time = GetCurrentTimestamp();
    int current_epoch = worker_get_epoch(MyDatabaseId);
    
    for (;;)
    {
        CHECK_FOR_INTERRUPTS();
        
        if (check_for_timeout(start_time))
            break;
        
        if (current_epoch != worker_get_epoch(MyDatabaseId))
            PG_RETURN_BOOL(true);
        
    }
    
    PG_RETURN_BOOL(false);
}
```

xuebinsu: Because we actually need to wait for 2 epoches instead of 1.

Consider the following sequence of operations (U stands for user session, B stands for bgworker):

1. B refresh_quota_model();
2. U: worker_get_epoch();
3. B: worker_increase_epoch();
4. U: worker_get_epoch();

Then wait_for_worker_new_epoch() will return with quota not refreshed at all.

xuebinsu: Besides, there is no `break;` but only a `NOTICE` on timeout. The user can decide whether to cancel the query manually.

Sasasu: > we actually need to wait for 2 epoches instead of 1.

I prefer to use `for(;;) {} for(;;) {}` rather than `for(;;){ for(;;) }`

> there is no break; but only a NOTICE on timeout.

still a dead loop, will lead to CI hang.
it makes no difference as before

quotamodel.c
```diff
@@ -26,6 +26,7 @@
 #include "executor/spi.h"
 #include "funcapi.h"
 #include "storage/ipc.h"
+#include "port/atomics.h"
```
beeender, line 29: not needed?

xuebinsu:
drop extension
timeout
set diskquota.worker_timeout
drop extension

create extension
timeout
set 
create extension

quotamodel.c
xuebinsu: It is needed. I use `pg_atomic_read()` and `pg_atomic_fetch_add()` to access the `epoch` so that the process only need to acquire `LW_SHARED` locks. This reudces lock contension significantly.

Sasasu (approved):
let us wait for #136
just ignore all code format problems in this PR

quotamodel.c
beeender: hmm, this is needed since the atomic usage is added to the `quotamodel.h`. Probably putting the header file there is better.
Minor issue.

beeender (approved):

PR #131, opened 2022-02-07
ereportif uses the 1st param as the condition

beeender:
Otherwise compiler reports a warning:
"comparison of constant ‘20’ with boolean expression is always false"


xuebinsu (approved):

xuebinsu (approved):

Sasasu (approved):
opps

PR #135, opened 2022-02-14
Add test cases for postmaster restart

beeender:
- When postmaster dies, the launcher and workers should be terminated.
- When postmaster restarts, they should be restarted as well.
- Move common ignored patterns to the top init file.


tests/isolation2/sql/test_postmaster_restart.sql
```diff
@@ -0,0 +1,46 @@
...
+
+-- stop postmaster
+!\retcode pg_ctl -D $MASTER_DATA_DIRECTORY -w stop;
+
+-- launcher should be terminated
+!\retcode pgrep -a -f "postgres.*launcher";
+-- bgworker should be terminated
+!\retcode pgrep -a -f "postgres.*diskquota.*isolation2test";
+
+-- start postmaster
```
beeender, line 25: `-E` should be changed to "-c gp_role=dispatch"
I will add a comment for this.

tests/isolation2/sql/test_postmaster_restart.sql
```diff
@@ -0,0 +1,46 @@
...
+
+-- launcher should exist
+!\retcode pgrep -a -f "postgres.*launcher";
+-- bgworker should exist
+!\retcode pgrep -a -f "postgres.*diskquota.*isolation2test";
+
+-- stop postmaster
+!\retcode pg_ctl -D $MASTER_DATA_DIRECTORY -w stop;
+
+-- launcher should be terminated
```
beeender, line 20: This will be flaky since it is not guaranteed the worker will be started at this point

beeender:
I will revisit this one soon

xuebinsu (approved):

tests/isolation2/sql/test_postmaster_restart.sql
```diff
@@ -0,0 +1,52 @@
+!\retcode gpconfig -c "diskquota.hard_limit" -v "on" > /dev/null;
```
higuoxing, line 1: Looks that all the operations in this file run synchronously. Maybe it fits in regression test?

beeender: nope, we have a postmaster restart during the test, which means the session will be disconnected. Regress test cannot handle it.

higuoxing: I think it can reconnect the database with `\c <dbname>`, am I wrong? I'm ok with this being an isolation test.

https://github.com/greenplum-db/gpdb/blob/83bb65fa979744bbc531b246524977aef2210668/src/test/regress/sql/bitmap_index.sql#L312

beeender: yeah, `\c` should work as well, tested

```
❯ psql testdb
psql (9.4.26)
Type "help" for help.

testdb=# select * from pg_backend_pid();
 pg_backend_pid
----------------
         267984
(1 row)

testdb=# select * from pg_backend_pid();
FATAL:  terminating connection due to administrator command
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Succeeded.
testdb=# \c testdb
You are now connected to database "testdb" as user "cc".
testdb=#
```

I am too lazy to change this to a regress test. Let's keep this as it is?

higuoxing (approved):

PR #136, opened 2022-02-14
Add clang-format and editorconfig

beeender:
- clang-format is from https://groups.google.com/a/greenplum.org/g/gpdb-dev/c/rHb4DjSd1iI/m/FKPbqwYcBAAJ
- See editorconfig at https://editorconfig.org/

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,40 @@
+---
+BasedOnStyle: Google
+
+# How much whitespace?
+UseTab: ForIndentation
+TabWidth: 4
+IndentWidth: 4
+ContinuationIndentWidth: 8
+
+SpacesBeforeTrailingComments: 1
+
+# Line things up
+AccessModifierOffset: -4  # outdent `public:`, etc
+
+DerivePointerAlignment: false
+PointerAlignment: Right   # char *foo, char &bar
+
+AlignConsecutiveAssignments: true
+AlignConsecutiveDeclarations: true
+
+# Braces
+AlwaysBreakAfterReturnType: TopLevelDefinitions
+AllowShortFunctionsOnASingleLine: Inline
+BreakBeforeBraces: Custom
+BraceWrapping:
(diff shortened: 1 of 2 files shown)


beeender:
Let me know if you have a better clang-format.

xuebinsu:
It is good to have a formatter. Can we automate this, e.g., force formatting on pushing to Github? It is hard to remember so many things when committing code.

Sasasu (approved):
should we format all existing before publishing a new version?

beeender:
> It is good to have a formatter. Can we automate this, e.g., force formatting on pushing to Github? It is hard to remember so many things when committing code.

We can either make a pre-git commit hook? any automation formatting after commit will change the SHA1 which is not accepted. We can have a formatted checker for CI.

beeender:
> should we format all existing before publishing a new version?

I am struggling with this, shall we?

xuebinsu:
Hi @beeender and @Sasasu ,

> We can either make a pre-git commit hook?
This sounds good. Then how can we ensure that the pre-commit hook is installed on everyone's machine? How about installing it before `make installcheck`?

> We can have a formatted checker for CI.
This sounds good too. Let's add the check in the scripts.

 > should we format all existing before publishing a new version?
I think so. I think we should avoid big changes after releasing a major version.

higuoxing:
> > We can either make a pre-git commit hook?
> 
> This sounds good. Then how can we ensure that the pre-commit hook is installed on everyone's machine? How about installing it before `make installcheck`?

A script called `git-clang-format` is shipped with `clang-format`. If you have installed `clang-format`, you can invoke it via

```bash
$ git clang-format <commit>
```

> > We can have a formatted checker for CI.
> 
> This sounds good too. Let's add the check in the scripts.

+1 to this idea.

xuebinsu (approved):

PR #138, opened 2022-02-16
Extend fetch_table_stat() to update db cache

xuebinsu:

This PR is to replace PR #99 , and depends on PR #130 to fix a race
condition that occurs after `CREATE EXTENSION`.

### DIFF
--- a/diskquota--1.0.sql
+++ b/diskquota--1.0.sql
@@ -21,11 +21,6 @@ RETURNS void STRICT
 AS 'MODULE_PATHNAME'
 LANGUAGE C;
 
-CREATE FUNCTION diskquota.update_diskquota_db_list(oid, int4)
-RETURNS void STRICT
-AS 'MODULE_PATHNAME'
-LANGUAGE C;
-
 CREATE TABLE diskquota.table_size (tableid oid, size bigint, PRIMARY KEY(tableid));
 
 CREATE TABLE diskquota.state (state int, PRIMARY KEY(state));
--- a/diskquota.c
+++ b/diskquota.c
@@ -129,14 +129,14 @@ _PG_init(void)
 	init_disk_quota_enforcement();
 	init_active_table_hook();
 
+	/* Add dq_object_access_hook to handle drop extension event. */
+	register_diskquota_object_access_hook();
+
 	/* start disk quota launcher only on master */
 	if (!IS_QUERY_DISPATCHER())
 	{
 		return;
 	}
-
-	/* Add dq_object_access_hook to handle drop extension event. */
-	register_diskquota_object_access_hook();
 
 	/* set up common data for diskquota launcher worker */
 	worker.bgw_flags = BGWORKER_SHMEM_ACCESS |
--- a/diskquota.c
+++ b/diskquota.c
@@ -523,10 +523,7 @@ create_monitor_db_table(void)
 	bool		ret = true;
 
 	sql = "create schema if not exists diskquota_namespace;"
-		"create table if not exists diskquota_namespace.database_list(dbid oid not null unique);"
-		"create schema if not exists diskquota;"
-		"create or replace function diskquota.update_diskquota_db_list(oid, int4) returns void "
-		"strict as '$libdir/diskquota' language C;";
+		"create table if not exists diskquota_namespace.database_list(dbid oid not null unique);";
 
 	StartTransactionCommand();
 
@@ -891,15 +888,6 @@ del_dbid_from_database_list(Oid dbid)
 	{
 		ereport(ERROR, (errmsg("[diskquota launcher] SPI_execute sql:'%s', errno:%d", str.data, errno)));
 	}
-	pfree(str.data);
-
-	/* clean the dbid from shared memory*/
-	initStringInfo(&str);
-	appendStringInfo(&str, "select gp_segment_id, diskquota.update_diskquota_db_list(%u, 1)"
-			" from gp_dist_random('gp_id');", dbid);
-	ret = SPI_execute(str.data, true, 0);
-	if (ret != SPI_OK_SELECT)
-		ereport(ERROR, (errmsg("[diskquota launcher] SPI_execute sql:'%s', errno:%d", str.data, errno)));
 	pfree(str.data);
 }
 
--- a/diskquota.h
+++ b/diskquota.h
@@ -32,7 +32,9 @@ typedef enum
 typedef enum
 {
 	FETCH_ACTIVE_OID,			/* fetch active table list */
-	FETCH_ACTIVE_SIZE			/* fetch size for active tables */
+	FETCH_ACTIVE_SIZE,			/* fetch size for active tables */
+	ADD_DB_TO_MONITOR,
+	REMOVE_MONITORED_DB,
 }			FetchTableStatType;
 
 typedef enum
--- a/diskquota.h
+++ b/diskquota.h
@@ -34,7 +34,7 @@ typedef enum
 	FETCH_ACTIVE_OID,			/* fetch active table list */
 	FETCH_ACTIVE_SIZE,			/* fetch size for active tables */
 	ADD_DB_TO_MONITOR,
-	REMOVE_MONITORED_DB,
+	REMOVE_DB_FROM_BEING_MONITORED,
 }			FetchTableStatType;
 
 typedef enum
--- a/diskquota_utility.c
+++ b/diskquota_utility.c
@@ -505,6 +505,11 @@ dq_object_access_hook(ObjectAccessType access, Oid classId,
 	 */
 	update_diskquota_db_list(MyDatabaseId, HASH_REMOVE);
 
+	if (!IS_QUERY_DISPATCHER())
+	{
+		return;
+	}
+
 	/*
 	 * Lock on extension_ddl_lock to avoid multiple backend create diskquota
 	 * extension at the same time.
--- a/diskquota_utility.c
+++ b/diskquota_utility.c
@@ -56,7 +56,6 @@ PG_FUNCTION_INFO_V1(set_schema_quota);
 PG_FUNCTION_INFO_V1(set_role_quota);
 PG_FUNCTION_INFO_V1(set_schema_tablespace_quota);
 PG_FUNCTION_INFO_V1(set_role_tablespace_quota);
-PG_FUNCTION_INFO_V1(update_diskquota_db_list);
 PG_FUNCTION_INFO_V1(set_per_segment_quota);
 PG_FUNCTION_INFO_V1(relation_size_local);
 
@@ -549,6 +548,7 @@ dq_object_access_hook(ObjectAccessType access, Oid classId,
 	}
 	LWLockRelease(diskquota_locks.extension_ddl_message_lock);
 	LWLockRelease(diskquota_locks.extension_ddl_lock);
+	update_diskquota_db_list(MyDatabaseId, HASH_REMOVE);
 out:
 	if (next_object_access_hook)
 		(*next_object_access_hook) (access, classId, objectId,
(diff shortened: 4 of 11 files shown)


diskquota.h
```diff
@@ -32,7 +32,9 @@ typedef enum
 typedef enum
 {
 	FETCH_ACTIVE_OID,			/* fetch active table list */
-	FETCH_ACTIVE_SIZE			/* fetch size for active tables */
+	FETCH_ACTIVE_SIZE,			/* fetch size for active tables */
+	ADD_DB_TO_MONITOR,
+	REMOVE_MONITORED_DB,
 }			FetchTableStatType;
```
Sasasu, line 38: I am checking what we do in `FetchTableStat` event loop.

Sasasu: The logic is fine, except the naming is a little confusing.

but I have some questions:

why there are so many 'copy to temp hash map and iterate the map to do something':
for example `remove_committed_relation_from_cache`

```c
void remove_committed_relation_from_cache(void) {
	HASH_SEQ_STATUS iter = {0};
	DiskQuotaRelationCacheEntry *entry = NULL;
	DiskQuotaRelationCacheEntry *local_entry = NULL;
	HTAB *local_relation_cache;
	HASHCTL	ctl;

	memset(&ctl, 0, sizeof(ctl));
	ctl.keysize = sizeof(Oid);
	ctl.entrysize = sizeof(DiskQuotaRelationCacheEntry);
	ctl.hcxt = CurrentMemoryContext;
	ctl.hash = oid_hash;

	local_relation_cache = hash_create("local relation cache", 1024, &ctl, HASH_ELEM | HASH_CONTEXT | HASH_FUNCTION);

	LWLockAcquire(diskquota_locks.relation_cache_lock, LW_SHARED);
	hash_seq_init(&iter, relation_cache);
	while ((entry = hash_seq_search(&iter)) != NULL)
	{
		local_entry = hash_search(local_relation_cache, &entry->relid, HASH_ENTER, NULL);
		memcpy(local_entry, entry, sizeof(DiskQuotaRelationCacheEntry));
	}
	LWLockRelease(diskquota_locks.relation_cache_lock);
	
	hash_seq_init(&iter, local_relation_cache);
	while ((local_entry = hash_seq_search(&iter)) != NULL) {
		if (OidIsValid(RelidByRelfilenode(local_entry->rnode.node.spcNode, local_entry->rnode.node.relNode)))
			remove_cache_entry(InvalidOid, local_entry->rnode.node.relNode);
	}
	hash_destroy(local_relation_cache);
}
```
can be

```c
void remove_committed_relation_from_cache(void) {
	HASH_SEQ_STATUS iter = {0};
	DiskQuotaRelationCacheEntry *entry = NULL;

	size_t to_remove_size = 0;
	DiskQuotaRelationCacheEntry *to_remove = NULL;

	LWLockAcquire(diskquota_locks.relation_cache_lock, LW_SHARED);
	to_remove_size = hash_get_num_entries(relation_cache);
	to_remove = palloc(sizeof(DiskQuotaRelidCacheEntry) * to_remove_size);

	size_t to_remove_index = 0;
	hash_seq_init(&iter, relation_cache);
	while ((entry = hash_seq_search(&iter)) != NULL) {
		memcpy(&to_remove[to_remove_index], entry, sizeof(DiskQuotaRelationCacheEntry));
		to_remove_index++;
	}
	LWLockRelease(diskquota_locks.relation_cache_lock);

	for (size_t i = 0; i < to_remove_index; ++i) {
		DiskQuotaRelationCacheEntry entry = to_remove[i];
		if (OidIsValid(RelidByRelfilenode(entry.rnode.node.spcNode, entry.rnode.node.relNode)))
			remove_cache_entry(InvalidOid, entry.rnode.node.relNode);
	}

	pfree(to_remove);
}
```

the second version is faster than the first and much easier to understand.

the same logic also appear in `gp_fetch_active_tables > pull_active_list_from_seg > convert_map_to_string` and many.

diskquota.c
```diff
@@ -872,16 +869,17 @@ del_dbid_from_database_list(Oid dbid)
...
-	pfree(str.data);
 
 	/* clean the dbid from shared memory*/
-	initStringInfo(&str);
-	appendStringInfo(&str, "select gp_segment_id, diskquota.update_diskquota_db_list(%u, 1)"
-			" from gp_dist_random('gp_id');", dbid);
+	resetStringInfo(&str);
+	appendStringInfo(&str, 
+			"select gp_segment_id, diskquota.diskquota_fetch_table_stat(%u, ARRAY[]::oid[])"
+			" from gp_dist_random('gp_id');", REMOVE_MONITORED_DB);
```
Sasasu, line 877: it will remove the dbid that diskquota launcher lived in.

Sasasu (approved):
this ~~confusing~~ API change looks good to me

diskquota.c
xuebinsu: Good point! Will fix that.

diskquota.h
beeender: nice finding, let's fix this in a separated PR>

quotamodel.c
```diff
@@ -623,13 +623,15 @@ do_check_diskquota_state_is_ready(void)
 
 	/* Add the dbid to watching list, so the hook can catch the table change*/
 	initStringInfo(&sql_command);
-	appendStringInfo(&sql_command, "select gp_segment_id, diskquota.update_diskquota_db_list(%u, 0) from gp_dist_random('gp_id')  UNION ALL select -1, diskquota.update_diskquota_db_list(%u, 0);",
-				MyDatabaseId, MyDatabaseId);
+	appendStringInfo(&sql_command,
```
beeender, line 626: Add a comment above:
```
// update_diskquota_db_list on segments
```

quotamodel.c
```diff
@@ -623,13 +623,15 @@ do_check_diskquota_state_is_ready(void)
...
+					"FROM gp_dist_random('gp_id');", ADD_DB_TO_MONITOR);
 	ret = SPI_execute(sql_command.data, true, 0);
-        if (ret != SPI_OK_SELECT)
-                ereport(ERROR, (errcode(ERRCODE_INTERNAL_ERROR),
-                                                errmsg("[diskquota] check diskquota state SPI_execute failed: error code %d", ret)));
+    if (ret != SPI_OK_SELECT)
+        ereport(ERROR, (errcode(ERRCODE_INTERNAL_ERROR),
+            	errmsg("[diskquota] check diskquota state SPI_execute failed: error code %d", ret)));
 	pfree(sql_command.data);
+	update_diskquota_db_list(MyDatabaseId, HASH_ENTER);
```
beeender, line 634: Add a comment above:
```
// update_diskquota_db_list on coordinator
```

gp_activetable.c
```diff
@@ -400,6 +400,12 @@ diskquota_fetch_table_stat(PG_FUNCTION_ARGS)
 			case FETCH_ACTIVE_SIZE:
 				localCacheTable = get_active_tables_stats(PG_GETARG_ARRAYTYPE_P(1));
 				break;
+			case ADD_DB_TO_MONITOR:
```
beeender, line 403: The function is doing more than the function comment says. Add a comment in the function header. Probably a TODO that we make a better function name later?

diskquota_utility.c
```diff
@@ -1055,7 +1046,7 @@ update_diskquota_db_list(PG_FUNCTION_ARGS)
 	}
 	LWLockRelease(diskquota_locks.monitoring_dbid_cache_lock);
 
-	PG_RETURN_VOID();
+	return found;
```
beeender, line 1049: Why do you even return a `bool` that nobody is checking the return value? Does the caller can do anything if it returns `false`? If not, keep the void return, and log a warning(?) to say which `dbid` cannot be found by action xxx

diskquota.h
```diff
@@ -32,7 +32,9 @@ typedef enum
 typedef enum
 {
 	FETCH_ACTIVE_OID,			/* fetch active table list */
-	FETCH_ACTIVE_SIZE			/* fetch size for active tables */
+	FETCH_ACTIVE_SIZE,			/* fetch size for active tables */
+	ADD_DB_TO_MONITOR,
+	REMOVE_MONITORED_DB,
```
beeender, line 37: To match this `ADD_DB_TO_MONITOR`, wouldn't `REMOVE_DB_FROM_MONITOR` be a better name?

gp_activetable.c
```diff
@@ -400,6 +400,12 @@ diskquota_fetch_table_stat(PG_FUNCTION_ARGS)
...
+				update_diskquota_db_list(MyDatabaseId, HASH_ENTER);
+				break;
+			case REMOVE_MONITORED_DB:
```
beeender, line 406: This is never called for now. Do we plan to call it somewhere? e.g., when bgworker quit, if we do have a plan, add a `TODO` of `FIXME` somewhere.

diskquota.h
xuebinsu: I think using the passive voice is more appropriate, like `REMOVE_DB_FROM_BEING_MONITORED`, but that would be too long.

gp_activetable.c
xuebinsu: It is used in `dq_object_access()`, which is called during `DROP EXTENSION`.

diskquota.h
beeender: Maybe simplely `ADD_MONITOR` `REMOVE_MONITOR`?

diskquota_utility.c
```diff
@@ -1027,7 +1033,7 @@ update_diskquota_db_list(Oid dbid, HASHACTION action)
 	if (action == HASH_ENTER)
 	{	
 		Oid *entry = NULL;
-		entry = hash_search(monitoring_dbid_cache, &dbid, HASH_ENTER, &found);
+		entry = hash_search(monitoring_dbid_cache, &dbid, HASH_ENTER_NULL, &found);
```
beeender, line 1041: If you want to use `HASH_ENTER_NULL`, then you need to check the returned `entry` and report an error. in this case, using `||` instead of `&&` at 1038 will be more accurate. Otherwise, we risk a null pointer issue.

```c
void *
hash_search_with_hash_value(HTAB *hashp,
							const void *keyPtr,
							uint32 hashvalue,
							HASHACTION action,
							bool *foundPtr)
{
...
	if (foundPtr)
		*foundPtr = (bool) (currBucket != NULL);
...
		case HASH_ENTER_NULL:
			/* FALL THRU */

		case HASH_ENTER:
}
```

xuebinsu: Will `hash_search()` return `NULL` if the entry is found?

diskquota.h
xuebinsu: Since "monitor" can also be a noun, using `ADD_MONITOR` and `REMOVE_MONITOR` will be very confusing. We are adding a DB, not a monitor after all.

Compared to `REMOVE_MONITOR`, I prefer `REMOVE_DB_FROM_BEING_MONITORED`.

xuebinsu: "monitor" here is a verb basically. `ADD_DB_TO_MONITOR` means "add db in order to monitor it".

diskquota_utility.c
beeender: > Will hash_search() return NULL if the entry is found?
No, I think?

But it could return `found = false` and `entry = NULL`

then `if (!found && entry == NULL)` will be `false`, but actually we have a `entry = NULL` for the later statements.

diskquota.h
beeender: Both are fine to me, as long as they look like a pair. :+1:

diskquota_utility.c
xuebinsu: Well, then how about change this to `if (entry == NULL)`?

diskquota--1.0.sql
```diff
@@ -21,11 +21,6 @@ RETURNS void STRICT
 AS 'MODULE_PATHNAME'
 LANGUAGE C;
 
-CREATE FUNCTION diskquota.update_diskquota_db_list(oid, int4)
```
beeender, line 24: According to our upgrade discussion, if we want to backport this to `1.x`, then we need to bump the minor version.

Also, if we don't do a backport, this should be removed from `diskquota--1.0.sql`. Let's discuss this when we implement the upgrade.

beeender (dismissed):

Sasasu (approved):

Sasasu (approved):

beeender (approved):

PR #139, opened 2022-02-16
Allow nonsuperuser roles to manange disk qouta

reshke:
The purpose of these patches is to allow non-superuser roles to managed  disk qouta in greeplum cluster. This is helpfully for cloud providers which can serve managed greenplum solution. This is undesirable to grant superuser to some customer roles, because superusers are able to perform almost anything on cluster VMs. The solution in to create some 'special' role (diskquota_admin) with needed permission and grant it to customer-side role

### DIFF
--- a/diskquota_utility.c
+++ b/diskquota_utility.c
@@ -339,11 +339,13 @@ dispatch_pause_or_resume_command(Oid dbid, bool pause_extension)
 Datum
 diskquota_pause(PG_FUNCTION_ARGS)
 {
-	if (!superuser())
+	Oid role;
+	role = get_role_oid("diskquota_admin", true);
+	if (!superuser() && !is_member_of_role(GetUserId(), role))
 	{
 		ereport(ERROR,
 				(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
-				 errmsg("must be superuser to pause diskquota")));
+				 errmsg("must be superuser or diskquota_admin to pause diskquota")));
 	}
 
 	Oid dbid = MyDatabaseId;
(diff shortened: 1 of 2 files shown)


xuebinsu:
Hi,

It will be great if we can allow non-`SUPERUSER`s to use Diskquota safely. 

However, granting the privileges of the UDFs to non-`SUPERUSER`s is currently unsafe. For example, if user A calls `diskquota.enable_hardlimit()`, it will affect user B.

Could you please elaborate more on how to avoid this?

Best regards.

x4m:
> However, granting the privileges of the UDFs to non-`SUPERUSER`s is currently unsafe. For example, if user A calls `diskquota.enable_hardlimit()`, it will affect user B.
> 
> Could you please elaborate more on how to avoid this?
Database administrator can simply avoid granting diskquota_admin to untrusted user.

reshke:
despite code changes there is another thing to do: we need to  actually create role diskqouta_admin in sql files to make things actually work. But i dont understand, should i add 
`
CREATE ROLE diskqouta_admin;
GRANT ALL ON SCHEMA diskquota to diskqouta_admin;
GRANT ALL ON TABLE diskquota.quota_config TO diskqouta_admin;
......
etc for all relations
`

to file diskquota--2.0.sql or diskquota--1.0.sql ?

beeender:
> in addition to code changes there is another thing to do: we need to actually create role diskqouta_admin in sql files to make things actually work. But i dont understand, should i add
CREATE ROLE diskqouta_admin; GRANT ALL ON SCHEMA diskquota to diskqouta_admin; GRANT ALL ON TABLE diskquota.quota_config TO diskqouta_admin; ...... etc for all relations

True, without the SQL change, this probably doesn't work out of box.

We are discussing another change that moves `quota_config` to the `diskquota` database to solve some annoying "table locking" issues when dropping the extension. Plus maybe add some utilities for quota management. In that case, the way of implementing the `diskquota_admin` might be a bit different. 

We will revisit this PR later.

PR #142, opened 2022-02-17
Use guc for hardlimit

beeender:
- Change to use GUC to set hardlimit instead of udf. Since the hardlimit
  setting needs to be persistent after postmaster restarting.
- Fix relevant test cases.

### DIFF
--- a/quotamodel.c
+++ b/quotamodel.c
@@ -166,8 +166,6 @@ static HTAB *table_size_map = NULL;
 static HTAB *disk_quota_black_map = NULL;
 static HTAB *local_disk_quota_black_map = NULL;
 
-pg_atomic_uint32 *diskquota_hardlimit = NULL;
-
 static shmem_startup_hook_type prev_shmem_startup_hook = NULL;
 
 /* functions to maintain the quota maps */
@@ -445,12 +443,6 @@ disk_quota_shmem_startup(void)
 			&hash_ctl,
 			HASH_ELEM | HASH_FUNCTION);
 
-	diskquota_hardlimit = ShmemInitStruct("diskquota_hardlimit",
-										  sizeof(pg_atomic_uint32),
-										  &found);
-	if (!found)
-		memset((void *) diskquota_hardlimit, 0, sizeof(pg_atomic_uint32));
-
 	/* use disk_quota_worker_map to manage diskquota worker processes. */
 	memset(&hash_ctl, 0, sizeof(hash_ctl));
 	hash_ctl.keysize = sizeof(Oid);
@@ -506,7 +498,6 @@ DiskQuotaShmemSize(void)
 	size = add_size(size, hash_estimate_size(MAX_NUM_MONITORED_DB, sizeof(Oid)));
 	size = add_size(size, hash_estimate_size(MAX_NUM_MONITORED_DB, sizeof(DiskQuotaWorkerEntry)));
 	size = add_size(size, hash_estimate_size(diskquota_max_active_tables, sizeof(Oid)));
-	size += sizeof(bool); /* sizeof(*diskquota_hardlimit) */
 	return size;
 }
 
@@ -745,7 +736,7 @@ refresh_disk_quota_usage(bool is_init)
 		/* copy local black map back to shared black map */
 		flush_local_black_map();
 		/* Dispatch blackmap entries to segments to perform hard-limit. */
-		if (pg_atomic_read_u32(diskquota_hardlimit))
+		if (diskquota_hardlimit)
 			dispatch_blackmap(local_active_table_stat_map);
 		hash_destroy(local_active_table_stat_map);
 	}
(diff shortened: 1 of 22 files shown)


quotamodel.c
```diff
@@ -2143,80 +2134,3 @@ show_blackmap(PG_FUNCTION_ARGS)
...
-		ereport(ERROR,
-				(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
-				 errmsg("must be superuser to enable hardlimit")));
-
-	/*
-	 * If this UDF is executed on segment servers, we should clear
-	 * the blackmap firstly, or the relation may be blocked by the
-	 * blackmap dispatched by the previous iteration.
-	 */
-	if (!IS_QUERY_DISPATCHER())
```
beeender, line 2188: @higuoxing Why do we need this before?

higuoxing: If we don't clear the blackmap on segment servers, consider the following situation:

1. Relation `t` exceeds some quota we set.
2. The blackmaps on segment servers are refreshed (Relation `t` is added to the blackmap, we cannot insert anything into `t`).
3. The diskquota is paused and the blackmaps on segment servers are not cleared. When the diskquota is paused, the blackmap on segment servers won't be updated.
4. Let's delete some data from relation `t` and perform `VACUUM FULL` on it to make the size of relation `t` smaller than the quota.
5. Resume the diskquota.
6. Before the blackmaps on segment servers being updated, the relation `t` still exists in blackmaps. We cannot insert data into relation `t` until the blackmaps get updated in the next refreshing iteration.

Q: Why not clear the blackmap on master server?
A: The blackmap on master server keeps being updated even if the diskquota is paused before 5ebc81d50068ee7ff8291393d90d158997729ebf. (See: https://github.com/greenplum-db/diskquota/blame/d04e5ef1bebe11534bed763786e49ea4bb4d686c/diskquota.c#L333)

tests/regress/expected/test_ctas_pause.out
```diff
@@ -1,11 +1,7 @@
 CREATE SCHEMA hardlimit_s;
 SET search_path TO hardlimit_s;
-SELECT diskquota.enable_hardlimit();
- enable_hardlimit 
-------------------
- 
-(1 row)
-
+\! gpconfig -c "diskquota.hard_limit" -v "true" > /dev/null
```
xuebinsu, line 3: I think it is more common to use "on"/"off" than "true"/"false".

xuebinsu (commented):
LGTM. Please re-trigger CI.

quotamodel.c
beeender: What about we clear the blackmap on segment whenever the `pause()` is dispatched? 
The above case looks more like a logic related with `pause()` but not hardlimit setting. I feel it is probably do a unique handling for `pause()`.

higuoxing: > What about we clear the blackmap on segment whenever the `pause()` is dispatched? The above case looks more like a logic related with `pause()` but not hardlimit setting. I feel probably doing unique handling in `pause()` is better.

Sounds reasonable to me.

xuebinsu (approved):

tests/regress/expected/test_ctas_tablespace_role.out
```diff
@@ -86,9 +82,5 @@ RESET default_tablespace;
 DROP TABLESPACE ctas_rolespc;
 REVOKE USAGE ON SCHEMA diskquota FROM hardlimit_r;
 DROP ROLE hardlimit_r;
-SELECT diskquota.disable_hardlimit();
- disable_hardlimit 
--------------------
- 
-(1 row)
-
+\! gpconfig -c "diskquota.hard_limit" -v "false" > /dev/null
```
xuebinsu, line 85: Why not change all of them?

quotamodel.c
beeender: OK, I will do the `pause()` part modification in a separated PR.

tests/regress/expected/test_ctas_tablespace_role.out
beeender: oops, how could it be missed. I did a `sed` ....

beeender: .... missed the false part .. 
do it now

quotamodel.c
xuebinsu: Not clearing the blackmap seems also reasonable to me. It is expected that the user needs to wait for a new epoch before the change to take effect.

xuebinsu (approved):

Sasasu (approved):

quotamodel.c
beeender: > Not clearing the blackmap seems also reasonable to me. It is expected that the user needs to wait for a new epoch before the change to take effect.

hmm, actually yes. If the blackmap gets cleared, user need to wait a new epoch before the hardlimit take effect (get the blackmap dispatched). I tend to agree we don't have to clear it, it is intended behaviour.

PR #115, opened 2021-12-23
[CI] Add PR pipeline for diskquota.

higuoxing:
This PR adds pr-pipeline for diskquota. Also, this a test PR for our
pipeline :-)

PR #143, opened 2022-02-18
    Use ytt to create pipelines

beeender:
- Rewrite the pipelines by using ytt for code reuse.
- Add `fly.sh` for easier manipulating with pipelines.
- Create configs for pr, commit and dev pipelines.
- Fix the test task for rhel8.
- Use build/test images pair for all distros. And modify the
  `build_diskquota.sh` so it won't need `Python.h` during the configure
  stage.
- Fix the pipiline bug which a `ubuntu` test may use a `rhel6` diskquota
  binary to do the test. That is caused by the same name of the task
  output.

To replace #115

### DIFF
--- a/concourse/scripts/build_diskquota.sh
+++ b/concourse/scripts/build_diskquota.sh
@@ -20,43 +20,30 @@ function pkg() {
     pushd /usr/local/greenplum-db-devel/
     echo 'cp -r lib share $GPHOME || exit 1'> install_gpdb_component
     chmod a+x install_gpdb_component
+    install_files=( \
+        "lib/postgresql/diskquota.so" \
+        "share/postgresql/extension/diskquota.control" \
+        "share/postgresql/extension/diskquota--1.0.sql" \
+        "share/postgresql/extension/diskquota--2.0.sql" \
+        "share/postgresql/extension/diskquota--1.0--2.0.sql" \
+        "share/postgresql/extension/diskquota--2.0--1.0.sql" \
+        "install_gpdb_component")
     case "$DISKQUOTA_OS" in
     rhel6)
         tar -czf $TOP_DIR/diskquota_artifacts/diskquota-${DISKQUOTA_VERSION}-rhel6_x86_64.tar.gz \
-        lib/postgresql/diskquota.so \
-        share/postgresql/extension/diskquota.control \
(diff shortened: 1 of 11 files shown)


beeender:
A testing dev pipeline can be seen at https://extensions.ci.gpdb.pivotal.io/teams/main/pipelines/DEV:mulong_diskquota/jobs/build_test/builds/8

xuebinsu (approved):
LGTM

concourse/scripts/build_diskquota.sh
```diff
@@ -20,43 +20,30 @@ function pkg() {
...
     chmod a+x install_gpdb_component
+    install_files=( \
+        "lib/postgresql/diskquota.so" \
+        "share/postgresql/extension/diskquota.control" \
+        "share/postgresql/extension/diskquota--1.0.sql" \
+        "share/postgresql/extension/diskquota--2.0.sql" \
+        "share/postgresql/extension/diskquota--1.0--2.0.sql" \
+        "share/postgresql/extension/diskquota--2.0--1.0.sql" \
+        "install_gpdb_component")
     case "$DISKQUOTA_OS" in
```
higuoxing, line 31: Looks that the variable `DISKQUOTA_OS` is set and immediately used in [concourse/pipeline/job_def.lib.yml](https://github.com/greenplum-db/diskquota/pull/143/files#diff-c09e583fd3d13aeb15ba4b34f59fa998910e4717884d1d8633277dfbc75a6bf9). Can we simplify this `case` statement with the following script?

```
tar -czf $TOP_DIR/diskquota_artifacts/diskquota-${DISKQUOTA_VERSION}-${DISKQUOTA_OS}_x86_64.tar.gz \
    "${install_files[@]}"
```

higuoxing (commented):
LGTM with one nit.

concourse/scripts/build_diskquota.sh
beeender: good idea, will change it

liuxueyang (approved):

concourse/scripts/build_diskquota.sh
beeender: I will create a separate PR for that

PR #144, opened 2022-02-21
Create worker entry before starting worker

xuebinsu:
Currently, the Diskquota launcher first starts a worker, then creates
the worker entry. However, after the worker starts, it cannot find the
entry when trying to check the is_paused status. Also, after a GPDB
restart, when the QD checks whether the worker is running by
checking the epoch, it might also fail to find the entry.

This patch fixes the issue by first creating the worker entry then
starting the bgworker process.

### DIFF
--- a/diskquota.h
+++ b/diskquota.h
@@ -113,7 +113,6 @@ typedef struct DiskQuotaWorkerEntry DiskQuotaWorkerEntry;
 struct DiskQuotaWorkerEntry
 {
 	Oid			dbid;
-	pid_t		pid;			/* worker pid */
 	pg_atomic_uint32 epoch; 		/* this counter will be increased after each worker loop */
 	bool is_paused; 			/* true if this worker is paused */
 	BackgroundWorkerHandle *handle;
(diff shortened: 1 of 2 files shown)


beeender:
close for testing ci. will reopen soon

diskquota.h
```diff
@@ -113,7 +113,6 @@ typedef struct DiskQuotaWorkerEntry DiskQuotaWorkerEntry;
 struct DiskQuotaWorkerEntry
 {
 	Oid			dbid;
```
liuxueyang, line 115: `DiskQuotaWorkerEntry.dbid` is not used as well. Not sure whether it's needed.

liuxueyang (approved):

diskquota.h
beeender: If it is not used, we can remove it in a separated PR

beeender (approved):

Sasasu (approved):

diskquota.h
xuebinsu: the `dbid` is used as key to the `HTAB`.

PR #147, opened 2022-02-21
Reset naptime when test finish

beeender:
Add tear_down to regress and isolation2 tests to reset the naptime. When
naptime is 0, quite a lot of cpu will be used. It will be a problem if
there are failing jobs on CI before getting reaped by the concourse. The
concourse worker CPU will be occupied which may impact following tests.


xuebinsu:
Actually, we can set naptime to a larger default value, like 10 (seconds), since our test does not depend on that.

Sasasu (dismissed):

beeender:
need to rename `tear_down` to `reset_config`

Sasasu (approved):

xuebinsu (approved):

PR #149, opened 2022-02-21
Fix miss-leading-indentation and missing pfree

beeender:
- pfree is not called when ereport error.
- gcc (GCC) 11.2.0 gives below warning:

```
quotamodel.c:621:5: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
  621 |     if (ret != SPI_OK_SELECT)
      |     ^~
quotamodel.c:624:9: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the ‘if’
  624 |         pfree(sql_command.data);
      |         ^~~~~# Please enter the commit message for your changes. Lines starting
```

PR #151, opened 2022-02-22
Extend fetch_table_stat() to update db cache v2

xuebinsu:

This PR is a revision of PR #138  , and depends on #130 to fix a race
condition that occurs after CREATE EXTENSION.

### DIFF
--- a/diskquota--1.0.sql
+++ b/diskquota--1.0.sql
@@ -21,11 +21,6 @@ RETURNS void STRICT
 AS 'MODULE_PATHNAME'
 LANGUAGE C;
 
-CREATE FUNCTION diskquota.update_diskquota_db_list(oid, int4)
-RETURNS void STRICT
-AS 'MODULE_PATHNAME'
-LANGUAGE C;
-
 CREATE TABLE diskquota.table_size (tableid oid, size bigint, PRIMARY KEY(tableid));
 
 CREATE TABLE diskquota.state (state int, PRIMARY KEY(state));
--- a/diskquota--1.0.sql
+++ b/diskquota--1.0.sql
@@ -21,6 +21,11 @@ RETURNS void STRICT
 AS 'MODULE_PATHNAME'
 LANGUAGE C;
 
+CREATE FUNCTION diskquota.update_diskquota_db_list(oid, int4)
+RETURNS void STRICT
+AS 'MODULE_PATHNAME'
+LANGUAGE C;
+
 CREATE TABLE diskquota.table_size (tableid oid, size bigint, PRIMARY KEY(tableid));
 
 CREATE TABLE diskquota.state (state int, PRIMARY KEY(state));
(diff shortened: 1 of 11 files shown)


liuxueyang (dismissed):

quotamodel.c
```diff
@@ -621,15 +621,18 @@ do_check_diskquota_state_is_ready(void)
...
+	/* Add current database to the monitored db cache on all segments */
+	appendStringInfo(&sql_command, 
+					"SELECT diskquota.diskquota_fetch_table_stat(%d, ARRAY[]::oid[]) "
+					"FROM gp_dist_random('gp_id');", ADD_DB_TO_MONITOR);
 	ret = SPI_execute(sql_command.data, true, 0);
-        if (ret != SPI_OK_SELECT)
-                ereport(ERROR, (errcode(ERRCODE_INTERNAL_ERROR),
-                                                errmsg("[diskquota] check diskquota state SPI_execute failed: error code %d", ret)));
+    if (ret != SPI_OK_SELECT)
+        ereport(ERROR, (errcode(ERRCODE_INTERNAL_ERROR),
```
beeender, line 631: I am going to close #149. Please just fix it here.

diskquota--1.0.sql
```diff
@@ -21,11 +21,6 @@ RETURNS void STRICT
 AS 'MODULE_PATHNAME'
 LANGUAGE C;
 
-CREATE FUNCTION diskquota.update_diskquota_db_list(oid, int4)
```
beeender, line 24: Since we re-work on this PR, let's keep the UDF for `1.x` for compatibility? So I don't have to add it back later for the backport work.

quotamodel.c
```diff
@@ -621,15 +621,20 @@ do_check_diskquota_state_is_ready(void)
...
+					"FROM gp_dist_random('gp_id');", ADD_DB_TO_MONITOR);
 	ret = SPI_execute(sql_command.data, true, 0);
-        if (ret != SPI_OK_SELECT)
-                ereport(ERROR, (errcode(ERRCODE_INTERNAL_ERROR),
-                                                errmsg("[diskquota] check diskquota state SPI_execute failed: error code %d", ret)));
+	if (ret != SPI_OK_SELECT) {
+		pfree(sql_command.data);
+		ereport(ERROR, (errcode(ERRCODE_INTERNAL_ERROR),
+				errmsg("[diskquota] check diskquota state SPI_execute failed: error code %d", ret)));
+    }
```
beeender, line 634: intend is wrong, use tab

beeender: ..... it might be me to make the spaces here ....

xuebinsu: I simply did a cherry-pick. I will change that.

beeender (approved):

liuxueyang (approved):

PR #154, opened 2022-02-23
Small tuning on logs

beeender:
Add logs when launcher/bgworker terminates
Also add missing infomation to the logs.


diskquota.c
```diff
@@ -342,8 +347,10 @@ disk_quota_worker_main(Datum main_arg)
 			usleep(1);
 
 		/* Emergency bailout if postmaster has died */
-		if (rc & WL_POSTMASTER_DEATH)
+		if (rc & WL_POSTMASTER_DEATH) {
```
beeender, line 350: wrong format, will fix

diskquota.c
```diff
@@ -342,8 +347,10 @@ disk_quota_worker_main(Datum main_arg)
...
+			ereport(LOG, (errmsg(term_postmaster_msg, dbname)));
```
xuebinsu, line 352: How about using `snprintf()`?

diskquota.c
```diff
@@ -622,13 +646,20 @@ start_workers_from_dblist(void)
 	PushActiveSnapshot(GetTransactionSnapshot());
 	ret = SPI_connect();
 	if (ret != SPI_OK_CONNECT)
-		ereport(ERROR, (errmsg("[diskquota launcher] SPI connect error, errno:%d", errno)));
+		ereport(ERROR, (errmsg("[diskquota launcher] SPI connect error, errno: %d, return code: %d.", errno, ret)));
```
xuebinsu, line 650: Can we have test case to ensure that these messages are actually logged rather than ignored in `PG_CATCH()`?

diskquota.c
```diff
@@ -622,13 +646,20 @@ start_workers_from_dblist(void)
...
 	if (ret != SPI_OK_SELECT)
-		ereport(ERROR, (errmsg("select diskquota_namespace.database_list")));
+        ereport(ERROR, (errmsg(
+                        "[diskquota launcher] 'select diskquota_namespace.database_list', errno: %d, return code: %d",
+                        errno, ret)));
 	tupdesc = SPI_tuptable->tupdesc;
 	if (tupdesc->natts != 1 || tupdesc->attrs[0]->atttypid != OIDOID)
-		ereport(ERROR, (errmsg("[diskquota launcher] table database_list corrupt, laucher will exit")));
+	{
+		ereport(LOG, (errmsg("[diskquota launcher], natts/atttypid: %d.",
```
xuebinsu, line 659: This message will be logged in each iteration. Does that matter?

diskquota.c
```diff
@@ -1005,13 +1039,13 @@ worker_set_handle(Oid dbid, BackgroundWorkerHandle *handle)
...
 		workerentry->handle = handle;
-	} 
+	}
 	LWLockRelease(diskquota_locks.worker_map_lock);
 	if (!found)
 	{
 		ereport(ERROR, (errcode(ERRCODE_INTERNAL_ERROR),
-						errmsg("[diskquota] worker not found for database \"%s\"",
-							   get_database_name(dbid))));	
+						errmsg("[diskquota] worker not found for database '%s'",
```
xuebinsu, line 1048: I think it is a convention to use double quotes:
```bash
[gpadmin@localhost diskquota]$ psql hello
psql: FATAL:  database "hello" does not exist
```

beeender: hmm, OK, will change that back

diskquota.c
beeender: it will only be logged if there is an error, the next statement is a `ereport(ERROR)`

Sasasu (approved):

diskquota.c
beeender: I removed the `const char*` and duplicate the log strings in every case just like normal. Which is much easier to read. Using `const char*` for log formatting purposes doesn't look good.

diskquota.c
beeender: done

diskquota.c
beeender: #155 @liuxueyang is working on that

beeender:
@xuebinsu re-review

xuebinsu (approved):

PR #155, opened 2022-02-24
Missing EmitErrorReport before FlushErrorState

beeender:
FlushErrorState() will clear the error state. To get the error log in
server logs, EmitErrorReport() needs to be called before flushing. All
other places with FlushErrorState() have the Emit before in the
codebase.

For diskquota bgworker, use `SearchSysCache1`. Because the bgworker
should be error-tolerant to keep it running in the background, we
can't throw an error. If we `EmitErrorReport` in bgworker, then the
results returned by `CdbDispatchCommand` would be incorrect. On the
other hand, the diskquota launcher can throw an error if needed.


Sasasu (dismissed):

higuoxing (dismissed):
LG

beeender:
CI fails, checking

beeender:
`EmitErrorReport` causes some timing changes which fails the `hard_limit` test.
And it would be the good time we separate the `diskquota_relation_open` for QD and bgworker. (bgworker can just use `try_relation_open()`)

xuebinsu:
Great!

diskquota.c
```diff
@@ -678,7 +679,9 @@ start_workers_from_dblist(void)
 	ret = SPI_connect();
 	if (ret != SPI_OK_CONNECT)
 		ereport(ERROR, (errmsg("[diskquota launcher] SPI connect error, errno:%d", errno)));
-	ret = SPI_execute("select dbid from diskquota_namespace.database_list;", true, 0);
+	sql = "select dbid from diskquota_namespace.database_list;";
+	debug_query_string = sql;
```
beeender, line 683: Don't forget remove this debug var

tests/regress/expected/test_query_string.out
```diff
@@ -0,0 +1,97 @@
...
+SELECT diskquota.wait_for_worker_new_epoch();
+ wait_for_worker_new_epoch 
+---------------------------
+ t
+(1 row)
+
+-- expect insert fail
+INSERT INTO a SELECT generate_series(1,100);
+ERROR:  schema's disk space quota exceeded with name:s1
+SELECT pg_sleep(1);
```
beeender, line 48: why `pg_sleep()` here? To wait for the log to be flushed to the file?

tests/regress/expected/test_query_string.out
```diff
@@ -0,0 +1,97 @@
...
+DROP EXTENSION diskquota;
+\c contrib_regression
+DROP DATABASE query_string_db;
+\! gpconfig -c log_statement -m 'all' -v 'none' > /dev/null
+\! gpstop -u > /dev/null
+\! gpconfig --show log_statement
+Values on all segments are consistent
+GUC          : log_statement
+Master  value: all
+Segment value: none
```
beeender, line 97: The way of checking logs looks good! :+1:

diskquota_utility.c
```diff
@@ -170,11 +171,13 @@ init_table_size_table(PG_FUNCTION_ARGS)
 	appendStringInfo(&buf,
 					 "update diskquota.state set state = %u;",
 					 DISKQUOTA_READY_STATE);
+	debug_query_string = buf.data;
```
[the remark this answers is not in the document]
beeender: It looks like the query string should be put in the log if SPI fails through the callback, see https://github.com/greenplum-db/gpdb/blob/83bb65fa979744bbc531b246524977aef2210668/src/backend/executor/spi.c#L2802

Doesn't it work? Or I misreading the code?

xuebinsu:
LGTM. Will +1 after CI turns green.

.clang-format
```diff
@@ -35,5 +35,6 @@ BraceWrapping:
   BeforeCatch: true
   SplitEmptyFunction: false
   SplitEmptyRecord: false
+  BeforeElse: true
```
xuebinsu, line 38: Is this intended?

xuebinsu (approved):
LGTM

.clang-format
liuxueyang: Yeah, we should use this style:

```
}
else
{
}
```

instead of

```
} else {
}
```

relation_cache.c
```diff
@@ -188,22 +188,31 @@ update_relation_cache(Oid relid)
...
+parse_primary_table_oid(Oid relid, bool on_bgworker)
 {
 	Relation rel;
 	Oid namespace;
 	Oid  parsed_oid;
 	char relname[NAMEDATALEN];
 
-	rel = diskquota_relation_open(relid, NoLock);
-	if (rel == NULL)
+	if (on_bgworker)
```
beeender, line 198: Add comments to say why do we use different routines for bgworker/non-bgworker.

quotamodel.c
```diff
@@ -1318,6 +1323,28 @@ get_rel_owner_schema_tablespace(Oid relid, Oid *ownerOid, Oid *nsOid, Oid *table
...
+	HeapTuple tp;
+
+	tp         = SearchSysCache1(RELOID, ObjectIdGetDatum(relid));
+	bool found = HeapTupleIsValid(tp);
+	if (HeapTupleIsValid(tp))
+	{
+		Form_pg_class reltup = (Form_pg_class)GETSTRUCT(tp);
+
+		*nsOid = reltup->relnamespace;
+		memcpy(relname, reltup->relname.data, NAMEDATALEN);
```
beeender, line 1341: This function copies into a given pointer argument and it could be error-prone for the future callers since the caller needs to know the `relname` is pre-malloced/array to pass to this function.

Do we plan to use this function somewhere else than `parse_primary_table_oid()`? If we do, we probably need to make the API looks better, at least a comment to say something about the `relname`.
Otherwise, just move the function body to the `parse_primary_table_oid()` or move it to `relation_cache.c` and make it static. Since the function itself doesn't have too much thing to do with `quotamodle.c`.

quotamodel.c
```diff
@@ -1318,6 +1323,28 @@ get_rel_owner_schema_tablespace(Oid relid, Oid *ownerOid, Oid *nsOid, Oid *table
...
+ * Given table oid, search for namespace and name.
+ */
+bool
+get_rel_name_namespace(Oid relid, Oid *nsOid, char *relname)
+{
+	HeapTuple tp;
+
+	tp         = SearchSysCache1(RELOID, ObjectIdGetDatum(relid));
+	bool found = HeapTupleIsValid(tp);
+	if (HeapTupleIsValid(tp))
```
beeender, line 1336: to `if (found)` is more readable IMO.

relation_cache.c
```diff
@@ -188,22 +188,31 @@ update_relation_cache(Oid relid)
...
 	{
-		return InvalidOid;
+		if (!get_rel_name_namespace(relid, &namespace, relname))
```
beeender, line 200: Just curious, what if `SearchSysCache1` throws? Be caught by https://github.com/greenplum-db/diskquota/pull/155/files#diff-37c3cd3a9010e9ee712fafe8387841a81e91db77c361f85fc366a2cab83de197L712 ?

beeender (commented):
Looks good, just some minor comments. But I prefer they can be fixed before merging. Let me if you have a different opinion.

quotamodel.c
liuxueyang: There is a similar function `get_rel_owner_schema_tablespace` doing similar things above in `quotamodel.c`. So I put it here. I'll add a comment for `*relname`.

beeender (commented):
LGTM

Since some commits are submitted by myself, I cannot do `+1` on GitHub.

tests/regress/expected/test_query_string.out
```diff
@@ -0,0 +1,122 @@
...
+
+SELECT DISTINCT ON (diskquota_related) logmessage AS diskquota_related
+FROM gp_toolkit.gp_log_system
+WHERE logmessage LIKE '%diskquota%'
+  AND logmessage LIKE '%pg_extension%'
+  AND logmessage NOT LIKE '%gp_toolkit%'
+  AND logtime >= NOW() - INTERVAL '5 minutes';
+                                           diskquota_related                                            
+--------------------------------------------------------------------------------------------------------
+ statement: -- complain if script is sourced in psql, rather than via CREATE EXTENSION                 +
```
Sasasu, line 75: there are too many contexts. I think this test is not useful for now.

and this test will fail after DDL modification.

gp_activetable.c
```diff
@@ -675,7 +679,7 @@ get_active_tables_oid(void)
 
 		if (relOid != InvalidOid)
 		{
-			prelid             = get_primary_table_oid(relOid);
+			prelid             = get_primary_table_oid(relOid, TRUE);
```
Sasasu, line 682: <stdbool.h> have `true` type. have better compiler integration.
is the time to abandon C89 bad style.

.clang-format
Sasasu: I think both is OK, but format code in normal feature PR is not a good idea.
squash to 2 commits, the first is code format, the second is add EmitErrorReport() would be better

tests/regress/expected/test_query_string.out
liuxueyang: Yeah, I think I can remove the `test_query_string` test case. @xuebinsu @beeender

Sasasu (approved):

fanfuxiaoran:
The pr title and description should be updated, as it looks like not only fixing the FlushErrorState.

.clang-format
liuxueyang: ok, I'll create another PR to add format changes.

liuxueyang:
> The pr title and description should be updated, as it looks like not only fixing the FlushErrorState.

updated commit message, please review.

```
FlushErrorState() will clear the error state. To get the error log in
server logs, EmitErrorReport() needs to be called before flushing. All
other places with FlushErrorState() have the Emit before in the
codebase.

For diskquota bgworker, use `SearchSysCache1`. Because the bgworker
should be error-tolerant to keep it running in the background, we
can't throw an error. If we `EmitErrorReport` in bgworker, then the
results returned by `CdbDispatchCommand` would be incorrect. On the
other hand, the diskquota launcher can throw an error if needed.
```

beeender:
LGTM :tada:

PR #158, opened 2022-02-25
[Backport] Extend fetch_table_stat() to update db cache v2 (#151)

beeender:

This PR is a revision of PR #138 , and depends on #130 to fix a race
condition that occurs after CREATE EXTENSION.

Compared to 9684b635, the test has been modified to use pg_sleep()
instead of wait_for_worker_new_epoch() since 1.x doesn't support it.

The extension created by 1.0.3 will still work with new the new so as
long as the UDF update_diskquota_db_list() is not called. (And it won't
be called in the new so.)

Co-authored-by: Chen Mulong <chenmulong@gmail.com>
(cherry picked from commit 9684b6350deb0a7d3849cf4cd202cd97c9b14882)

### DIFF
--- a/diskquota--1.0.sql
+++ b/diskquota--1.0.sql
@@ -21,6 +21,9 @@ RETURNS void STRICT
 AS 'MODULE_PATHNAME'
 LANGUAGE C;
 
+-- This UDF won't be used. But we keep it here so that it would be the same with
+-- 1.0.3. Easier for 2.x upgrade design. Be notice that, the C function is not
+-- there anymore, calling this UDF will report an error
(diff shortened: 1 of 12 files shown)


beeender:
CI is not working. Close this one and create a new one to try.

beeender:
PR pipeline is not working. See the CI results here:
https://extensions.ci.gpdb.pivotal.io/teams/main/pipelines/DEV:mulong_diskquota/jobs/build_test/builds/14

xuebinsu (approved):

Sasasu (approved):

PR #161, opened 2022-02-28
build: use cmake as build system

Sasasu:
this is PR only contains one minimal change.

the target is to optimize the upgrade process,  I will split it into many PR.


MINOR_VERSION
```diff
@@ -0,0 +1 @@
+1.0
```
beeender, line 1: instead of having three file: `MAJOR_VERSION`, `MINOR_VERSION` and `PATCH_VERSION`, i like to have a single `VERSION` file. Parse the version string is not a difficult thing to do in cmake. https://github.com/bytemaster/tornet/blob/master/CMakeModules/VersionMacros.cmake

Makefile
```diff
@@ -1,18 +1,3 @@
-# contrib/diskquota/Makefile
```
beeender, line 1: Are we going to make cmake totally work in this PR? Maybe we can just remove this file?

concourse/scripts/build_diskquota.sh
```diff
@@ -10,18 +10,23 @@ function pkg() {
...
+    if [ "${DISKQUOTA_OS}" = "rhel6" ]; then
+        export CC="$(which gcc)"
+    fi
+
     export USE_PGXS=1
     pushd diskquota_src/
     DISKQUOTA_VERSION=$(git describe --tags)
-    make clean
-    make install
+    mkdir build
```
beeender, line 20: use `diskquota_artifacts` instead of `build`. Then we don't have to change the pipeline, it will be passed to the next task automatically. And our pipeline can be compatible with `1.x`. Also added a comments here to say why do we use the  `diskquota_artifacts` here.

concourse/scripts/build_diskquota.sh
```diff
@@ -10,18 +10,23 @@ function pkg() {
...
+    cmake -B build .
+    make -C build install
     popd
 
     pushd /usr/local/greenplum-db-devel/
     echo 'cp -r lib share $GPHOME || exit 1'> install_gpdb_component
     chmod a+x install_gpdb_component
     tar -czf "$TOP_DIR/diskquota_artifacts/diskquota-${DISKQUOTA_VERSION}-${DISKQUOTA_OS}_x86_64.tar.gz" \
```
beeender, line 28: then we don't have to tar here.
we can just use cpack

diskquota.c
```diff
@@ -41,6 +41,7 @@ PG_MODULE_MAGIC;
 #define DISKQUOTA_DB	"diskquota"
 #define DISKQUOTA_APPLICATION_NAME  "gp_reserved_gpdiskquota"
 
+#include <unistd.h> // for useconds_t
```
beeender, line 44: interesting, it doesn't report error before?

CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+# generate version
+if(NOT DEFINED DISKQUOTA_MINOR_VERSION)
+  file(STRINGS "MINOR_VERSION" DISKQUOTA_MINOR_VERSION)
+endif()
+
+set(CMAKE_C_FLAGS
+    "${CMAKE_C_FLAGS} '-DDISKQUOTA_MINOR_VERSION=\"${DISKQUOTA_MINOR_VERSION}\"'"
+)
+
+if("${DISKQUOTA_MINOR_VERSION}" STREQUAL "1.0")
```
beeender, line 27: We don't need cmake support for 1.0 I think?

CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+  quotamodel.c
+  relation_cache.c)
+
+list(
+  APPEND
+  diskquota_DDL
+  diskquota.control
+  diskquota--1.0.sql
+  diskquota--1.0--2.0.sql
+  diskquota--2.0.sql
```
beeender, line 49: This needs to be removed as `diskquota-2.x`
And then we do a rename when `install()` at line 62.

beeender: no, i was wrong, this is good. the name should be fixed

Makefile
beeender: OK. The scope of this PR is to build with cmake only. This is good

CMakeLists.txt
beeender: Talked with @Sasasu . We keep it here for easier debugging. Plus this PR didn't bump the version, it is still 1.0 for so file.

diskquota.c
Sasasu: not report any error before.

makefile: `/usr/bin/clang -Wall -Wmissing-prototypes -Wpointer-arith -Wendif-labels -Wmissing-format-attribute -Wformat-security -fno-strict-aliasing -fwrapv -Wno-unused-but-set-variable -Wno-address -Wno-unused-command-line-argument -g -ggdb -ggdb -fno-inline -O0 -g3 -std=gnu99   -Werror=uninitialized -Werror=implicit-function-declaration -fPIC -I/home/sa/GPDB/install/include -I. -I./ -I/home/sa/GPDB/install/include/postgresql/server -I/home/sa/GPDB/install/include/postgresql/internal -D_GNU_SOURCE -I/usr/include/libxml2   -c -o diskquota.o diskquota.c`

cmake: `/usr/bin/cc -Ddiskquota_EXPORTS -I/home/sa/GPDB/install/include/postgresql/server -I/home/sa/GPDB/install/include -Wall -Wmissing-prototypes -Wpointer-arith -Wendif-labels -Wmissing-format-attribute -Wformat-security -fno-strict-aliasing -fwrapv -Wno-unused-but-set-variable -Wno-address -Wno-unused-command-line-argument -g -ggdb -ggdb -fno-inline -O0 -g3 -std=gnu99 -Werror=uninitialized -Werror=implicit-function-declaration -I/home/sa/GPDB/install/include       -DDISKQUOTA_VERSION="1.0.3"     -DDISKQUOTA_MAJOR_VERSION="1"   -DDISKQUOTA_MINOR_VERSION="0"   -DDISKQUOTA_PATCH_VERSION="3" -fPIC -std=gnu99 -MD -MT CMakeFiles/diskquota.dir/diskquota.c.o -MF CMakeFiles/diskquota.dir/diskquota.c.o.d -o CMakeFiles/diskquota.dir/diskquota.c.o -c /home/sa/GPDB/diskquota/diskquota.c`

concourse/scripts/build_diskquota.sh
Sasasu: will do this when doing pipeline refactor.  but for now, let's limit the scope.

xuebinsu (approved):

concourse/scripts/build_diskquota.sh
Sasasu: image
I can't remove this part from cpack output.

can not keep the compatibility :(

https://cmake.org/cmake/help/latest/module/CPack.html
> Note that the DESTINATION option of the [install()](https://cmake.org/cmake/help/latest/command/install.html#command:install) command must be a relative path; otherwise installed files are ignored by CPack.

then cpack install those files with `--prefix=diskquota-0.1.1-Linux` then zip the whole package dir.

xuebinsu (approved):

CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+  diskquota--2.0--1.0.sql)
+
+add_library(diskquota MODULE ${diskquota_SRC})
+
+# pg_config do not give us the `--prefix` we use `../../` as a workaround
+if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+  set(CMAKE_INSTALL_PREFIX "${PG_PKG_LIB_DIR}/../../")
```
beeender, line 68: Can we use `PG_HOME`?

https://github.com/greenplum-db/diskquota/pull/161/files#diff-c00eda21c7b337eca6f678990d9c036beba4a9e371a189eb8186651c05876257R39

beeender (approved):
One minor comment :+1:

CMakeLists.txt
Sasasu: will change to PG_HOME in next PR

concourse/scripts/build_diskquota.sh
beeender: Do this:

```
cpack -G TGZ -D CPACK_INCLUDE_TOPLEVEL_DIRECTORY=off
```
https://gitlab.kitware.com/jiangyilism/cmake/-/blob/master/Source/CPack/cmCPackSTGZGenerator.cxx#L22
@Sasasu

higuoxing:
Do we need to update the build instruction in README? https://github.com/greenplum-db/diskquota/blob/gpdb/README.md#install

PR #162, opened 2022-03-01
rename library file to diskquota-<major.minor>.so

Sasasu:
this PR introduces a new packing method. for now, everyone can create a package without CI by running
```
cmake ../ -DDISKQUOTA_PREVIOUS_INSTALLER=~/diskquota-1.0.3-rhel8_x86_64.tar.gz
```
image

and this PR renames the current library file to `diskquota-2.0.so`, this will break the old `gpconfig -v shared_preload_libraries` command, please note.

the next step is:
 - finish upgrade test (https://github.com/greenplum-db/diskquota/commit/d54ab08ec4d1d166bf9049af25b6190c689494a1)
 - create a package on CI (need modify all pipeline, download previous release)
 - add catalog check (install 1.0.x, do upgrade to 2.0.0, check UDF is same as install from 2.0.0)
 - add many checks to prevent ABI or API breaks.


diskquota--1.0--2.0.sql
```diff
@@ -1,115 +1,127 @@
...
-CREATE TYPE diskquota.blackmap_entry_detail AS
-  (target_type text, target_oid oid, database_oid oid,
-   tablespace_oid oid, seg_exceeded boolean, dbnode oid, spcnode oid, relnode oid, segid int);
-
-CREATE FUNCTION diskquota.show_blackmap()
-RETURNS setof diskquota.blackmap_entry_detail
-AS 'MODULE_PATHNAME', 'show_blackmap'
-LANGUAGE C;
+ALTER TABLE diskquota.table_size DROP CONSTRAINT table_size_pkey;
+ALTER TABLE diskquota.table_size ADD COLUMN segid smallint DEFAULT 0;
```
xuebinsu, line 12:
```suggestion
ALTER TABLE diskquota.table_size ADD COLUMN segid smallint DEFAULT -1; -- "-1" means the coordinator
```

diskquota.control
```diff
@@ -1,5 +1,5 @@
 # diskquota extension
 comment = 'Disk Quota Main Program'
 default_version = '2.0'
-module_pathname = '$libdir/diskquota'
+module_pathname = '$libdir/diskquota-do-not-use-this.so'
```
beeender, line 4: >    The value of this parameter will be substituted for each occurrence of MODULE_PATHNAME in the script file(s). If it is not set, no substitution is made. Typically, this is set to $libdir/shared_library_name and then MODULE_PATHNAME is used in CREATE FUNCTION commands for C-language functions, so that the script files do not need to hard-wire the name of the shared library.

To avoid ommited module path in the `CREATE FUNCTION`? This is good :+1:

diskquota--1.0.sql
```diff
@@ -1,74 +1,71 @@
...
-        pg_roles as pgr
+	pg_class as pgc,
+	diskquota.quota_config as qc,
+	pg_roles as pgr
 WHERE pgc.relowner = qc.targetoid and pgc.relowner = pgr.oid and ts.tableid = pgc.oid
 GROUP BY pgc.relowner, pgr.rolname, qc.quotalimitMB;
 
 CREATE VIEW diskquota.show_fast_database_size_view AS
-SELECT ((SELECT SUM(pg_relation_size(oid)) FROM pg_class WHERE oid <= 16384)+ (SELECT SUM(size) FROM diskquota.table_size)) AS dbsize;
+select (
```
beeender, line 60: what SQL formatter are u using? why mixed using upper/lower cases?

Sasasu: by hand :(

xuebinsu: How about using https://github.com/darold/pgFormatter ?

diskquota.control
xuebinsu: Do we still need `module_pathname`?

cmake/Distro.cmake
```diff
@@ -0,0 +1,39 @@
...
+
+if(NOT DISTRO_NAME)
+    SET(DISTRO_NAME unknown)
+    if(EXISTS "/etc/redhat-release")
+        file(READ /etc/redhat-release rh_release)
+        string(REGEX MATCH "CentOS release 6.*" matched6 "${rh_release}")
+        string(REGEX MATCH "CentOS Linux release 7.*" matched7 "${rh_release}")
+        string(REGEX MATCH "Red Hat Enterprise Linux release 8.*" matched_rhel8 "${rh_release}")
+        string(REGEX MATCH "CentOS Linux release 8.*" matched_centos8 "${rh_release}")
+        if (matched6)
```
xuebinsu, line 18: This is a little bit confusing. How about using `is_centos6`, `is_centos7`, etc.?

xuebinsu (commented):
LGTM overall

cmake/current_binary_name
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+cd "$(dirname "$0")"
+
+if [ $(grep -P '^1.0' ../VERSION) ]; then
+    echo -n "diskquota.so"
+else
+    echo -n "diskquota-$(grep -o -P '^\d+.\d+' ../VERSION).so"
```
xuebinsu, line 8: How about using `diskquota-$(cat VERSION)`?

beeender (commented):
Looks good. One question about the SQL changes, is there any real changes in the SQL except formatting? I probably need to carefully compare the files.

diskquota--1.0.sql
beeender: tried, `pgFormatter` looks good

Sasasu:
```
> echo '\\echo Use "CREATE EXTENSION diskquota" to load this file. \\quit' | pg_format
\echo Use "CREATE EXTENSION diskquota" to load this file.
quit
```

seems there is no option to change the behavior :(

xuebinsu: What behavior do you expect?

xuebinsu (approved):

diskquota--1.0.sql
```diff
@@ -1,74 +1,71 @@
...
-AS 'MODULE_PATHNAME'
-LANGUAGE C;
+CREATE FUNCTION diskquota.set_schema_quota(text, text) RETURNS void STRICT AS '$libdir/diskquota.so' LANGUAGE C;
+CREATE FUNCTION diskquota.set_role_quota(text, text) RETURNS void STRICT AS '$libdir/diskquota.so' LANGUAGE C;
+CREATE FUNCTION diskquota.update_diskquota_db_list(oid, int4) RETURNS void STRICT AS '$libdir/diskquota.so' LANGUAGE C;
+CREATE FUNCTION diskquota.init_table_size_table() RETURNS void STRICT AS '$libdir/diskquota.so' LANGUAGE C;
+CREATE FUNCTION diskquota.diskquota_fetch_table_stat(int4, oid[]) RETURNS setof diskquota.diskquota_active_table_type AS '$libdir/diskquota.so', 'diskquota_fetch_table_stat' LANGUAGE C VOLATILE;
 
 CREATE VIEW diskquota.show_fast_schema_quota_view AS
 select pgns.nspname as schema_name, pgc.relnamespace as schema_oid, qc.quotalimitMB as quota_in_mb, sum(ts.size) as nspsize_in_bytes
```
beeender, line 41: Change to the upper case together with others in the file? Since this is for 1.0. Not changing it is fine with me. I will copy this file to `1.x` branch when this is merged.

diskquota--2.0.sql
```diff
@@ -1,93 +1,121 @@
...
-  SELECT * FROM diskquota.show_blackmap() AS BM;
-
-CREATE TABLE diskquota.table_size (tableid oid, size bigint, segid smallint, PRIMARY KEY(tableid, segid)) DISTRIBUTED BY (tableid, segid);
-
-CREATE TABLE diskquota.state (state int, PRIMARY KEY(state)) DISTRIBUTED BY (state);
-
-INSERT INTO diskquota.state SELECT (count(relname) = 0)::int  FROM pg_class AS c, pg_namespace AS n WHERE c.oid > 16384 and relnamespace = n.oid and nspname != 'diskquota';
+-- diskquota.quota_config and diskquota.target is dump-able, other table can be generate on fly
+SELECT pg_catalog.pg_extension_config_dump('diskquota.quota_config', '');
+SELECT gp_segment_id, pg_catalog.pg_extension_config_dump('diskquota.quota_config', '') from gp_dist_random('gp_id');
```
beeender, line 35: `from` to `FROM`

diskquota--2.0.sql
```diff
@@ -1,93 +1,121 @@
...
+SELECT pg_catalog.pg_extension_config_dump('diskquota.target', '');
+SELECT gp_segment_id, pg_catalog.pg_extension_config_dump('diskquota.target', '') from gp_dist_random('gp_id');
```
beeender, line 37: ditto

diskquota--2.0.sql
```diff
@@ -107,12 +135,21 @@ from diskquota.table_size as ts,
 WHERE pgc.relowner = qc.targetoid and pgc.relowner = pgr.oid and ts.tableid = pgc.oid and qc.quotatype=1 and ts.segid=-1
 GROUP BY pgc.relowner, pgr.rolname, qc.quotalimitMB;
 
+CREATE VIEW diskquota.show_fast_database_size_view AS
+select (
```
beeender, line 139: upper cases from line 139 - 143

diskquota--2.0.sql
```diff
@@ -107,12 +135,21 @@ from diskquota.table_size as ts,
...
+    (select sum(pg_relation_size(oid)) from pg_class where oid <= 16384)
+        +
+    (select sum(size) from diskquota.table_size where segid = -1)
+) AS dbsize;
+
+CREATE VIEW diskquota.blackmap AS SELECT * FROM diskquota.show_blackmap() AS BM;
+
 CREATE VIEW diskquota.show_fast_schema_tablespace_quota_view AS
 select pgns.nspname as schema_name, pgc.relnamespace as schema_oid, pgsp.spcname as tablespace_name, pgc.reltablespace as tablespace_oid, qc.quotalimitMB as quota_in_mb, sum(ts.size) as nspsize_tablespace_in_bytes
 from diskquota.table_size as ts,
```
beeender, line 149: upper case `from`

beeender (dismissed):
Just merge this. I will submit a PR to fix the upper cases issue in the SQL

beeender (approved):

xuebinsu (approved):

PR #165, opened 2022-03-03
Install the latest cmake on CI

beeender:
We need the archive extrace feature from cmake 3.18.
Before the new version landing to the test/build images, install them in
our ci scripts.
A cmake binary has been uploaded to the concourse gcs for faster
downloading.

### DIFF
--- a/concourse/pipeline/res_def.yml
+++ b/concourse/pipeline/res_def.yml
@@ -102,7 +102,7 @@ resources:
     repository: gcr.io/data-gpdb-public-images/gpdb6-ubuntu18.04-test
     tag: latest
 
-#! gpdb binary on gcs is located as different folder for different version
+# gpdb binary on gcs is located as different folder for different version
 - name: bin_gpdb6_centos6
   type: gcs
   source:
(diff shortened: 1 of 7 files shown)


beeender:
The PR pipeline is expected to fail since it has not been flown yet. Please review first, then I can fly the new pipeline.

Sasasu (approved):

Sasasu:
76c840b6417309b518276a83b70640300ae43084

concourse/pipeline/res_def.yml
```diff
@@ -127,3 +127,11 @@ resources:
...
     json_key: ((concourse-gcs-resources-service-account-key))
     versioned_file: 6X_STABLE/bin_gpdb_rhel8/bin_gpdb.tar.gz
+
+# Other dependencies
+- name: bin_cmake
+  type: gcs
+  source:
+    bucket: gpdb-extensions-concourse-resources
+    json_key: ((extensions-gcs-service-account-key))
+    regexp: dependencies/cmake-(.*)-linux-x86_64.sh
```
xuebinsu, line 137: How about fixing a version here, in case someone changes the file on GCS unexpectedly?

beeender: it is intended to use the `(.*)` here, so we always got the latest version, easy to update cmake. and the cmake public API is quite stable to my experience, never meet a backward compatibility issue before.

beeender: And we'd love to remove this i we can have cmake in the image. will try to ask rel to see if they can help add it to the build/test image.

beeender: > `How about fixing a version her

There is a standard way to fix the resource version in concourse, so if we have a compatibility issue, the version can still be chose through the concourse UI.

xuebinsu (approved):

PR #169, opened 2022-03-05
mc/cmake installcheck

beeender:
CMake installcheck:

- Regress.cmake is introduced to add some helper functions which can
  easily create regress/isolation2 test targets in the CMakeLists.
- When create regress targets, cmake will generate a series of
  diff_<target>_<test> target to show result diff on a specific target.
- Porting the isolation2_regress build from Makefile to cmake.
- Moved the magic trap regress diff from the concourse script to cmake.
  If the test target is called with env `SHOW_REGRESS_DIFF=1`, then the
  regress.diff will be printed when the case fails.
- `sql`/`expected` will be linked to the cmake binary directory and run
  test from there. So the `results` won't pollute the source directory.

Concourse scripts restructure:

This commit also try to explore the way that how the concourse scripts
are structured. Currently, the container is started with root user, and
this cannot be changed in a short time. But we require gpadmin user
basically for everything.
An `entry.sh` is introduced as a simple skeleton for the scripts. It
make sure a gpadmin user is created and everything needed can be
accessed by the gpadmin with a permanent path. So we don't have too much
ENV magic in the task script any more. Most of the setup work should be
done in the `entry.sh` and the task script should only do a few simple
steps with absolute paths.
By doing this, when we hijack to the concourse container, login with
gpadmin user, environment should be ready. Simple copy & paste from
the task script will just work.

`upgrade_extension.sh` may not be needed anymore, and it won't work. It
will be cleaned up later when we have the full implementation of
upgrade.


tests/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/regress/regress_init_file
+    SQL_DIR ${CMAKE_CURRENT_SOURCE_DIR}/regress/sql
+    EXPECTED_DIR ${CMAKE_CURRENT_SOURCE_DIR}/regress/expected
+    DATA_DIR ${CMAKE_CURRENT_SOURCE_DIR}/data
+    SCHEDULE_FILE ${CMAKE_CURRENT_SOURCE_DIR}/regress/diskquota_schedule
+    REGRESS_OPTS
+    --load-extension=gp_inject_fault
+    --dbname=contrib_regression)
+
+RegressTarget_Add(isolation2
```
Sasasu, line 15: 👀 I will add upgrade test later

concourse/scripts/build_diskquota.sh
```diff
@@ -15,30 +10,13 @@ function pkg() {
...
-    tar -czf "$TOP_DIR/diskquota_artifacts/diskquota-${DISKQUOTA_VERSION}-${DISKQUOTA_OS}_x86_64.tar.gz" \
-        lib/postgresql/diskquota*.so \
-        "share/postgresql/extension/diskquota.control" \
-        "share/postgresql/extension/diskquota--1.0.sql" \
-        "share/postgresql/extension/diskquota--2.0.sql" \
-        "share/postgresql/extension/diskquota--1.0--2.0.sql" \
-        "share/postgresql/extension/diskquota--2.0--1.0.sql" \
-        "install_gpdb_component"
+    pushd /home/gpadmin/diskquota_artifacts
+    cmake /home/gpadmin/diskquota_src
```
Sasasu, line 14: need to add `-DDISKQUOTA_PREVIOUS_INSTALLER` in here to run the upgrade test.
👀

Sasasu (approved):

concourse/scripts/build_diskquota.sh
beeender: Will do this in another PR.

xuebinsu (approved):

PR #172, opened 2022-03-07
Pack the last released so in the pipeline

beeender:
Due to our upgrade design, the newly released diskquota should contain
all the so file of every latest minor release. The packaging system has
been implemented by cmake DISKQUOTA_PREVIOUS_INSTALLER.
This commit get the latest release through gcs regexp and pass it to the
build script.

The pipeline has been split into different jobs based on the OS type
because of:
- Easier to use the common resource name for build_diskquota.yml
- job_def.lib.yml becomes shorter
- Can span jobs into different concourse workers, which may bring some
  better performance.
- Running flaky test again is easier.

Also a gate job has been added which is the main entrance for all
pipelines. In the future, we may want to have different gate job for
different pipelines, like a clang-format checking gate for the PR
pipeline.

### DIFF
--- a/concourse/fly.sh
+++ b/concourse/fly.sh
@@ -99,7 +99,6 @@ set -v
     -c "${yml_path}" \
     -l "${workspace}/gp-continuous-integration/secrets/gpdb_common-ci-secrets.yml" \
     -l "${workspace}/gp-continuous-integration/secrets/gp-extensions-common.yml" \
-    -l "${workspace}/gp-continuous-integration/secrets/gpdb_common-ci-secrets.yml" \
     -l "${workspace}/gp-continuous-integration/secrets/gpdb_6X_STABLE-ci-secrets.prod.yml" \
     -v "diskquota-branch=${branch}"
 set +v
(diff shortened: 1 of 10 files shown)


concourse/scripts/build_diskquota.sh
```diff
@@ -11,7 +11,9 @@ function pkg() {
     fi
 
     pushd /home/gpadmin/diskquota_artifacts
-    cmake /home/gpadmin/diskquota_src
+    local last_release_diskquota
+    last_release_diskquota=$(readlink -e /home/gpadmin/last_released_diskquota_bin/diskquota-*.tar.gz)
+    cmake /home/gpadmin/diskquota_src -DDISKQUOTA_PREVIOUS_INSTALLER="${last_release_diskquota}"
```
xuebinsu, line 16:
```suggestion
    cmake /home/gpadmin/diskquota_src -DDISKQUOTA_LAST_RELEASE="${diskquota_last_release}"
```
Would this be more clear?

beeender: changed to `DISKQUOTA_LAST_RELEASE_PATH`

xuebinsu (approved):
beautiful and neat idea!

Sasasu (approved):

concourse/fly.sh
```diff
@@ -100,6 +100,7 @@ set -v
     -l "${workspace}/gp-continuous-integration/secrets/gpdb_common-ci-secrets.yml" \
     -l "${workspace}/gp-continuous-integration/secrets/gp-extensions-common.yml" \
     -l "${workspace}/gp-continuous-integration/secrets/gpdb_common-ci-secrets.yml" \
```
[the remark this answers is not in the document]
beeender: nope, just a stupid mistake i made.
fixed

Sasasu (approved):

PR #173, opened 2022-03-08
cmake: also link RESULTS_DIR to working dir

Sasasu:
my workflow:
- write many new tests.
- find -name '*.sql' | sed 's/sql/out/g' | xargs touch
- run those tests.
- check the output, if satisfied
- cp from results to expected

if `expected` and `result` in the same folder will be better.

### DIFF
--- a/cmake/Regress.cmake
+++ b/cmake/Regress.cmake
@@ -4,6 +4,7 @@
 # RegressTarget_Add(<name>
 #   SQL_DIR <sql_dir>
 #   EXPECTED_DIR <expected_dir>
+#   RESULTS_DIR <results_dir>
 #   [INIT_FILE <init_file_1> <init_file_2> ...]
 #   [SCHEDULE_FILE <schedule_file_1> <schedule_file_2> ...]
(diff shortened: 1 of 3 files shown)


beeender:
Shouldn't the `results` dir always exist in the `build` dir (`CMAKE_CURRENT_BINARY') since it is part of the intermediate build results?

For your requirement, I am thinking of another way:
In the `Regress.cmake` we can create some targets:


```
            add_custom_target(cp_res_exp_target_xxx
                COMMAND
                cp
                ${working_DIR}/results/${casename}.out || exit 0
                ${working_DIR}/expected/${casename}.out
                COMMAND
                echo ${working_DIR}/expected/${casename}.out
                COMMAND
                echo ${working_DIR}/results/${casename}.out
                )

```

With the shell completion, it would be easy to do the copy action.
But if the link way fits your workflow better, I am fine with it as well.

beeender (approved):

liuxueyang (approved):

Sasasu:
xueyang also meets this problem, I am merging this to prevent more people wasting on it :>.

PR #174, opened 2022-03-08
Remove Makefiles

beeender:
Everything has been moved to cmake.


Sasasu (approved):
please also remove `upgrade_test/sql/downgrade_extension.sql` and `upgrade_test/sql/upgrade_extension.sqll`
i recovered these files by mistake while rebasing.

higuoxing:
Doesn't relate to this PR. What do you think of keeping a `Makefile` in the root dir and invoke the `cmake` commands, so that we can `make && make install && make installcheck` like before ... I'm a cmake noob :cry: (though I will learn it someday).

xuebinsu (approved):

beeender:
> Doesn't relate to this PR. What do you think of keeping a Makefile in the root dir and invoke the cmake commands, so that we can make && make install && make installcheck like before ... I'm a cmake noob cry (though I will learn it someday).
Xuebin mentioned that yesterday:

> Xuebin
> Would it be more convenient to have such a Makefile in the root directory?
> .PHONY install
> install:
>    cmake -B build && make -C build/ install

> Mulong
> I am fine with adding a Makefile with some phony targets for convenience. But I wouldn't prefer to doc it in the README to advertise it as the "official" way to build the project.
Knowing the details of how cmake works would have some benefit, especially it is not such a difficult thing.
e.g.:
If we support gpdb7 and postgres with the same source tree, then we could do this:
mkdir build_gpdb6
cmake .. --DPG_CONFIG=/to/gpdb6/bin/pg_config
mkdir build_gpdb7
cmake .. --DPG_CONFIG=/to/gpdb7/bin/pg_config
mkdir build_pg
cmake .. --DPG_CONFIG=/to/pg/bin/pg_config
The three build can exist at the same time (even can run the installcheck in the same terminal session without source greenplum_path.sh with a little bit tweak in the CMakelists.txt), and this is hard to achieve with Makefile.
So, feel free to create a PR for the root Makefile for dev's convenience, but keep the cmake details in the README

>Xuebin
One advantage of having a consistent build-and-run interface is to reduce the time to bootstrap a new-comer.

So I am fine to have it if it is useful for others. Feel free to add it. (Actually I prefer the cmake way, no pollution on the source tree, and it is not that hard to get used to.)

PR #175, opened 2022-03-08
Fix gcc warning

beeender:
Reported by gcc 11.2.0:

quotamodel.c:2115:25: warning: ‘strncpy’ output truncated before
terminating nul copying 10 bytes from a string of the same length
[-Wstringop-truncation]
 2115 |                         strncpy(targettype_str, "ROLE_QUOTA", 10);
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/quotamodel.c
+++ b/quotamodel.c
@@ -2116,16 +2116,16 @@ show_blackmap(PG_FUNCTION_ARGS)
 			StrNCpy(targettype_str, "ROLE_QUOTA", _TARGETTYPE_STR_SIZE);
 			break;
 		case NAMESPACE_QUOTA:
-			strncpy(targettype_str, "NAMESPACE_QUOTA", _TARGETTYPE_STR_SIZE);
+			StrNCpy(targettype_str, "NAMESPACE_QUOTA", _TARGETTYPE_STR_SIZE);
 			break;
 		case ROLE_TABLESPACE_QUOTA:
-			strncpy(targettype_str, "ROLE_TABLESPACE_QUOTA", _TARGETTYPE_STR_SIZE);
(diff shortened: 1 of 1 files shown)


Sasasu (approved):

quotamodel.c
```diff
@@ -2112,19 +2113,19 @@ show_blackmap(PG_FUNCTION_ARGS)
 		switch ((QuotaType) keyitem.targettype)
 		{
 		case ROLE_QUOTA:
-			strncpy(targettype_str, "ROLE_QUOTA", 10);
+			StrNCpy(targettype_str, "ROLE_QUOTA", _TARGETTYPE_STR_SIZE);
```
Sasasu, line 2116: opps

Sasasu (changes requested):

Sasasu (approved):

quotamodel.c
beeender: more opps!

Sasasu (approved):
`StrNCpy` is macro from postgresql 😂

liuxueyang (approved):

xuebinsu (approved):

quotamodel.c
higuoxing: If we use `StrNCpy`, I think we can remove the line `memset(targettype_str, 0, sizeof(targettype_str));` above the `switch` keyword, since `StrNCpy` set the rest of the buffer to `'\0'`. :rofl:

higuoxing: Besides, I found an interesting thread on the pgsql-hackers list (https://www.postgresql.org/message-id/flat/44f5e198-36f6-6cdb-7fa9-60e34784daae%402ndquadrant.com).

PR #177, opened 2022-03-09
Upload every build to GCS

beeender:
As versioned file.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -16,7 +16,7 @@ indent_size = 4
 indent_style = space
 indent_size = 4
 
-[*.{dxl,mdp}]
+[{*.cmake,CMakeLists.txt}]
 indent_style = space
 indent_size = 2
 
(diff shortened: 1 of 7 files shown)


beeender:
See testing pipeline here
https://extensions.ci.gpdb.pivotal.io/teams/main/pipelines/DEV:mulong_diskquota

xuebinsu:
Since we cannot use VERSION here, can we add a file that contains the latest commit hash when doing `tar`  to identify our dev build?

beeender:
No need, easy to know the info from the concourse for every build 
image

Sasasu (approved):

Sasasu:
I prefer to embed the build info in the binary if the build info is relay needed.
image

beeender:
> I prefer to embed the build info in the binary if the build info is relay needed.

It is doable. Any benefit by embedding into a binary file?
The current text file is mostly for our internal usage. e.g.: PM downloaded a binary from pipeline and found a problem. Then we can based on the text file in the tarball to identify where the binary was built from.

Sasasu (approved):
LGTM, the build info will not be copied to GPDB

xuebinsu (approved):

PR #178, opened 2022-03-10
Use cmake way to add C macro defines

beeender:
Otherwise it creates noise with LSP since the '-Dxxxx' is started with a
'\n\t'. And that cannot be parsed by ccls.


Sasasu (approved):
magic!

Sasasu (approved):

Sasasu (approved):

Sasasu (approved):

xuebinsu (approved):

PR #183, opened 2022-03-16
Format the code and add clang-format to PR pipeline

beeender:
- Use clang 13 to format the code
- Rename the job 'gate' to 'entrance', prepare adding 'exit' job
- Add clang-format as a entrance check into the PR pipeline

### DIFF
--- a/diskquota.c
+++ b/diskquota.c
@@ -249,9 +249,7 @@ disk_quota_worker_main(Datum main_arg)
 {
 	char *dbname = MyBgworkerEntry->bgw_name;
 
-	ereport(LOG, (errmsg("[diskquota] start disk quota worker process to "
-	                     "monitor database:%s",
-	                     dbname)));
+	ereport(LOG, (errmsg("[diskquota] start disk quota worker process to monitor database:%s", dbname)));
 
 	/* Establish signal handlers before unblocking signals. */
 	pqsignal(SIGHUP, disk_quota_sighup);
@@ -289,9 +287,7 @@ disk_quota_worker_main(Datum main_arg)
 
 		if (has_error)
 		{
-			static char _errfmt[] =
-			        "find issues in pg_class.pg_extension check server log. "
-			        "waited %d seconds",
+			static char _errfmt[] = "find issues in pg_class.pg_extension check server log. waited %d seconds",
 			            _errmsg[sizeof(_errfmt) + sizeof("2147483647" /* INT_MAX */) + 1] = {};
 			snprintf(_errmsg, sizeof(_errmsg), _errfmt, times * diskquota_naptime);
 
@@ -303,10 +299,8 @@ disk_quota_worker_main(Datum main_arg)
 		}
 
 		ereportif(!has_error && times == 0, WARNING,
-		          (errmsg("[diskquota] worker for \"%s\" detected the "
-		                  "installed version is \"%d.%d\", "
-		                  "but current version is %s. abort due to version not "
-		                  "match",
+		          (errmsg("[diskquota] worker for \"%s\" detected the installed version is \"%d.%d\", "
+		                  "but current version is %s. abort due to version not match",
 		                  dbname, major, minor, DISKQUOTA_VERSION),
 		           errhint("run alter extension diskquota update to \"%d.%d\"", DISKQUOTA_MAJOR_VERSION,
 		                   DISKQUOTA_MINOR_VERSION)));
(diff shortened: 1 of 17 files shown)


diskquota.c
```diff
@@ -15,15 +15,15 @@
...
-#include "gp_activetable.h"
 
 #include "postgres.h"
 
-#include "funcapi.h"
 #include "access/xact.h"
 #include "cdb/cdbvars.h"
 #include "commands/dbcommands.h"
 #include "executor/spi.h"
+#include "funcapi.h"
```
Sasasu, line 25: I'm worried the header has some `import order error`. so I prefer
```
diff --git a/.clang-format b/.clang-format
index 72ce535..061f06d 100644
--- a/.clang-format
+++ b/.clang-format
@@ -34,7 +34,4 @@ BraceWrapping:
   SplitEmptyFunction: false
   SplitEmptyRecord: false

-# Put "postgres.h" and "postgres_undefs.h" first in a group of includes.
-IncludeCategories:
-- Regex: '"postgres(_undefs)?.h"'
-  Priority: 1
+IncludeBlocks: Preserve
--
```

diskquota.c
```diff
@@ -444,18 +427,20 @@ disk_quota_worker_main(Datum main_arg)
 		 * background process goes away immediately in an emergency.
 		 */
 		rc = WaitLatch(&MyProc->procLatch,
-					   WL_LATCH_SET | WL_TIMEOUT | WL_POSTMASTER_DEATH,
-					   diskquota_naptime * 1000L);
+		               WL_LATCH_SET | WL_TIMEOUT | WL_POSTMASTER_DEATH,
+		               diskquota_naptime * 1000L);
 		ResetLatch(&MyProc->procLatch);
 
-		// be nice to scheduler when naptime == 0 and diskquota_is_paused() == true
```
Sasasu, line 451: `ColumnLimit: 120`?

diskquota.c
beeender: Actually the order issue happened to me yesterday, see https://github.com/greenplum-db/diskquota/pull/183/files/81bccd9738f7140cb0af181469bf17c9d5379dcf#diff-e9b3a87492b7d3b01934f6d0ed39419cbc07d90e68fa75d78959d51fa90db60fR4

add to `postgres.h` to the `.clang-format` looks good, will add it

beeender: set the `postgres.h` as 1st priority actually caught some header issue, see https://github.com/greenplum-db/diskquota/pull/183/files/81bccd9738f7140cb0af181469bf17c9d5379dcf#diff-e9b3a87492b7d3b01934f6d0ed39419cbc07d90e68fa75d78959d51fa90db60fR4

The header file uses some macro that is not included in the same header file.
But yeah, maybe we don't quite need this magic. Let me remove that rule for now.

beeender: nope, that rule is a good one. Otherwise, the `postgres.h` will be re-ordered to some unknown places, then the include orders is totally messed up and doesn't even compile.

Sasasu: I prefer this include orders:
1. our header file
1. PostgreSQL header file
1. <cdb_*.h>
1. third-party header file (if exist)
1. the standard C header
1. <sys/*.h>

I try to do this in 23472c4e, but give up. checking this in CI is best.
(I think clang-format is not smart to do this, the header file naming is chaos)

Sasasu (approved):

diskquota.c
beeender: my screen is normally split vertically into 3 windows, each one is 7:9 plus elder people loves 80 chars

diskquota.c
Sasasu: > nope, that rule is a good one. Otherwise, the `postgres.h` will be re-ordered to some unknown places, then the include orders is totally messed up and doesn't even compile.

for example `foo.c` file:

```
#include <postgres.h>

#include <foo.h>

// code start
```

foo.h
```
extern Oid bar_id;
```

with that rule, `foo.h` will not rise error but actually, it is not self-contain.

self-contain and less include are important to C++, it can significantly reduce the compile time.
for the C project, I just want `ccls` not raise error :(

beeender: clang-format is not good enough to apply this kind of rule. I think I prefer to DISABLE the include reorder for now. What do you think?

xuebinsu (approved):

Sasasu (approved):

PR #186, opened 2022-03-18
Support release build

beeender:
- Introduce macro DISKQUOTA_DEBUG through CMAKE_BUILD_TYPE. For the
  debug build, check the macro and allow naptime = 0. The min naptime
  for release build is 1 to prevent high cpu usage in user's
  environment.
- Add EXCLUDE arg to Regree.cmake to skip some tests for the release
  build since the release GPDB build doesn't support fault injector.
- Use GPDB release candidate to compile diskquota release. The release
  candidate has assertion disabled in the pg_config.h.
- Set naptime=0 in the config.sql won't take effect for diskquota
  release since it is less than the min value. So the naptime stays the
  same (normally the default value, 2). This fails some of the test,
  especially the hardlimit related ones. If the insertion takes less
  time than naptime (2 seconds, normally), it fails since the limitation
  has not been dispatched before finish. Thus, the naptime is set to 2
  seconds instead of 10 in the reset_config.

### DIFF
--- a/concourse/pipeline/res_def.yml
+++ b/concourse/pipeline/res_def.yml
@@ -111,30 +111,58 @@ resources:
     password: ((extensions-gcs-service-account-key))
 
 # gpdb binary on gcs is located as different folder for different version
+# Latest build with assertion enabled:
+# --enable-cassert --enable-tap-tests --enable-debug-extensions
+- name: bin_gpdb6_centos6_debug
+  type: gcs
+  source:
+    bucket: ((gcs-bucket))
+    json_key: ((concourse-gcs-resources-service-account-key))
+    regexp: server/published/gpdb6/server-rc-(.*)-rhel6_x86_64.debug.tar.gz
+- name: bin_gpdb6_centos7_debug
+  type: gcs
+  source:
+    bucket: ((gcs-bucket))
+    json_key: ((concourse-gcs-resources-service-account-key))
+    regexp: server/published/gpdb6/server-rc-(.*)-rhel7_x86_64.debug.tar.gz
+- name: bin_gpdb6_rhel8_debug
+  type: gcs
+  source:
+    bucket: ((gcs-bucket))
+    json_key: ((concourse-gcs-resources-service-account-key))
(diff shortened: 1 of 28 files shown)


beeender:
https://extensions.ci.gpdb.pivotal.io/teams/main/pipelines/DEV:mulong_diskquota
Testing pipeline for release build

concourse/pipeline/res_def.yml
```diff
@@ -111,30 +111,58 @@ resources:
...
+    regexp: server/release-candidates/gpdb6/greenplum-db-server-(.*)-centos7.04.tar.gz
+- name: bin_gpdb6_rhel8
   type: gcs
   source:
-    bucket: ((gcs-bucket-intermediates))
+    bucket: ((gcs-bucket))
     json_key: ((concourse-gcs-resources-service-account-key))
-    versioned_file: 6X_STABLE/bin_gpdb_ubuntu18.04/bin_gpdb.tar.gz
-- name: bin_gpdb6_rhel8
+    regexp: server/release-candidates/gpdb6/greenplum-db-server-(.*)-rhel8.04.tar.gz
```
Sasasu, line 159: dose rhel has version 8.04? and centos7.04?

beeender: uh..... very much like a problem with neovim. blame the editor

Sasasu (approved):

tests/CMakeLists.txt
```diff
@@ -1,5 +1,20 @@
 include(${CMAKE_SOURCE_DIR}/cmake/Regress.cmake)
 
+# Test cases need fault injector needs to be excluded from release build.
+# GPDB release build doesn't support fault injector
+list(APPEND exclude_regress_for_release
+    test_fetch_table_stat)
```
xuebinsu, line 6: How about moving this case to the isolation test?

xuebinsu (approved):
LGTM

tests/CMakeLists.txt
beeender: Why? Using fault inject is not a strong enough reason to move it I think? Plus it is not difficult to exclude it from release testing.

PR #190, opened 2022-03-22
Don't format errno as '%d' in error messages.

higuoxing:
Actually, errno *isn't* strictly an integer. It depends on your
operating system. Instead of formatting it as an integer, we should
transform it to C style string and format it as '%s'.

e.g., on my laptop the definition of errno is:

```C
/* The error code set by various library functions.  */
extern int *__errno_location (void) __THROW __attribute_const__;
# define errno (*__errno_location ())
```

### DIFF
--- a/diskquota.c
+++ b/diskquota.c
@@ -595,8 +595,8 @@ create_monitor_db_table(void)
 		ret_code = SPI_execute(sql, false, 0);
 		if (ret_code != SPI_OK_UTILITY)
 		{
-			ereport(ERROR, (errmsg("[diskquota launcher] SPI_execute error, sql: \"%s\", errno: %d, ret_code: %d.", sql,
-			                       errno, ret_code)));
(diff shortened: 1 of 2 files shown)


Sasasu (approved):

xuebinsu:
How will the `errno` be set during `SPI_execute()`?

Sasasu (approved):
in the old system, before pthread came along.
`errno` is a static i32 variable. but after the invention of thread or pthread, static i32 is not thread-safe. so `errno` becomes a function call, the function returns an i32 variable (as a lvalue, not rvalue).

this happens on everyone's machine, not only you.

beeender (approved):
I am not quite sure how useful `errno` is when SPI fails, but, better than nothing :)

beeender:
What is the behaviour of `strerror(0)` when the `errno` is not set?

Sasasu:
> What is the behavior of `strerror(0)` when the `errno` is not set?

"Success"

image

PR #191, opened 2022-03-24
Fix cmake build on Mac

beeender:
- Mac needs extra linker options that the symbols in the postgres
  executable can be found.
- Mac's grep doesn't have '-P' option.
- To make a extension so file, we use a 'module' target instead of
  'shared' on purpose. Since "A SHARED library may be marked with
  the FRAMEWORK target property to create an macOS Framework."
  Thus, the global link flags should be set through
  CMAKE_MODULE_LINKER_FLAGS than CMAK_SHARED_LINKER_FLAGS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,10 +30,12 @@ set(CMAKE_C_FLAGS_RELEASE "" CACHE
 set(CMAKE_C_FLAGS_DEBUG "-DDISKQUOTA_DEBUG"
     CACHE STRING "Flags for DEBUG build" FORCE)
 # set link flags for all sub-projects
-set(CMAKE_SHARED_LINKER_FLAGS "${PG_LD_FLAGS}")
+set(CMAKE_MODULE_LINKER_FLAGS "${PG_LD_FLAGS}")
+if (APPLE)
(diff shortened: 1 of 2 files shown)


liuxueyang (approved):

Sasasu (approved):

CMakeLists.txt
```diff
@@ -30,10 +30,12 @@ set(CMAKE_C_FLAGS_RELEASE "" CACHE
 set(CMAKE_C_FLAGS_DEBUG "-DDISKQUOTA_DEBUG"
     CACHE STRING "Flags for DEBUG build" FORCE)
 # set link flags for all sub-projects
-set(CMAKE_SHARED_LINKER_FLAGS "${PG_LD_FLAGS}")
+set(CMAKE_MODULE_LINKER_FLAGS "${PG_LD_FLAGS}")
+if (APPLE)
```
fanfuxiaoran, line 34: "APLLE" is interesting

fanfuxiaoran (approved):

CMakeLists.txt
beeender: https://cmake.org/cmake/help/latest/variable/APPLE.html

It is documented! XD

PR #192, opened 2022-03-25
Fix leak

beeender:
More commit message comes later
And there might still be one more leak


beeender:
Close this. @xuebinsu will create a new PR to fix other leaks together.

PR #197, opened 2022-03-31
fix test case for test_manytable

Sasasu:
rename manytable test to activetable_limit

    activetable_limit used to test if activetable_limit works.

    if number of current activetable > max_active_tables. bgworker will
    print some warring message. GPDB should not crash, but diskquota's basic
    function does not guarantee.


beeender (approved):
hmmm, `git mv` doesn't work?

xuebinsu (approved):
Good progress!

tests/regress/expected/test_activetable_limit.out
```diff
@@ -0,0 +1,38 @@
...
+CREATE EXTENSION diskquota;
+CREATE SCHEMA s;
+SELECT diskquota.set_schema_quota('s', '1 MB');
+ set_schema_quota 
+------------------
+ 
+(1 row)
+
+CREATE TABLE s.t1(i int) DISTRIBUTED BY (i);
+INSERT INTO s.t1 SELECT generate_series(1,100000); -- expect failed. diskquota should works. activetable = 1
```
Sasasu, line 24: not  failed

PR #198, opened 2022-03-31
Add the ytt based release pipeline

beeender:
The release pipeline will:
- Automatically make release build for every commit.
- The final release step needs to be started manually.
  - Then the build will be pushed to the release bucket.
  - And a tag will be pushed to the repository.


beeender:
Pipeline can be seen at

https://extensions.ci.gpdb.pivotal.io/teams/main/pipelines/RELEASE:diskquota

concourse/fly.sh
```diff
@@ -73,6 +73,17 @@ case ${pipeline_config} in
       pipeline_name="DEV:${pipeline_name}"
       config_file="dev.yml"
     ;;
+  release)
+      if [ -z "${pipeline_name}" ]; then
+          pipeline_name="RELEASE:diskquota"
```
beeender, line 78: Probably we use the name `RELEASE:diskquota:gpdb` is better, to align with the commit pipeline? I think I am going to change this if no objection.

beeender, line 78: Probably we use the name `RELEASE:diskquota:gpdb` is better, to align with the commit pipeline? I think I am going to change this if no objection.

concourse/pipeline/trigger_def.lib.yml
```diff
@@ -54,14 +58,44 @@ on_error:
...
+auto_trigger: true
+to_get:
+- get: diskquota_src
+  resource: diskquota_commit
+to_put_pre: #@ []
+to_put_post:
+- put: diskquota_commit
+  params:
+    repository: diskquota_src
+    tag: diskquota_src/VERSION
```
Sasasu, line 87: is this a placeholder? I don't find the logic for the version(tag) generate.

Sasasu (approved):

concourse/pipeline/trigger_def.lib.yml
beeender: It is black magic from concourse git resource 

> tag: Optional. If this is set then HEAD will be tagged. The value should be a path to a file containing the name of the tag.

https://github.com/concourse/git-resource

xuebinsu (approved):

PR #200, opened 2022-04-01
Use on/off in diskquota.status()

Sasasu:
PostgreSQL use on/off for all bool variable.

soft limits: on (soft will works) / paused (soft limit not works)
hard limits: on (hard will works) / off (hard limit not works) / paused (hard limit not works)

### DIFF
--- a/diskquota.c
+++ b/diskquota.c
@@ -1181,10 +1181,10 @@ diskquota_status_check_soft_limit()
 	LWLockRelease(diskquota_locks.worker_map_lock);
 
 	// if worker no booted, aka 'CREATE EXTENSION' not called, diskquota is paused
-	if (!found) return "paused";
+	if (!found) return "off";
 
 	// if worker booted, check 'worker_map->is_paused'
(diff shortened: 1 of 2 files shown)


xuebinsu (approved):

diskquota.c
```diff
@@ -1206,9 +1206,9 @@ diskquota_status_check_hard_limit()
 
 	// if worker booted and 'worker_map->is_paused == true' and hardlimit is enabled
 	// hard limits should also paused
-	if (found && paused && hardlimit) return "paused";
+	if (found && paused && hardlimit) return "off";
```
beeender, line 1209: if no `paused` is needed, shall we just return `off` here? Or you prefer to keep the code incase we need `paused` again?

beeender (commented):
no `paused` anymore? @radarwave

diskquota.c
Sasasu: `paused` is back.

radarwave (approved):

PR #201, opened 2022-04-07
Use released gpdb instead of release candidate

beeender:
By the gcs resource regex rule, 6.99.99 will be matched as the most
recent gpdb release candidate, which is not what we want. It is a
testing version from release team. And due to the go regex
implementation, (!?) is not supported to filter out an exact string.

So the release pipeline is changed to use the most recent RELEASED gpdb
binary instead.

### DIFF
--- a/concourse/fly.sh
+++ b/concourse/fly.sh
@@ -118,9 +118,10 @@ if [ "${pipeline_config}" == "dev" ]; then
     exit 0
 fi
 
+concourse_url=$(fly targets | awk "{if (\$1 == \"${target}\") {print \$2}}")
 echo ""
 echo "================================================================================"
 echo "Remeber to set the the webhook URL on GitHub:"
-echo "https://<target>.ci.gpdb.pivotal.io/api/v1/teams/main/pipelines/${pipeline_name}/resources/${hook_res}/check/webhook?webhook_token=<hook_token>"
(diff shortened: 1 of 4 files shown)


beeender:
Another change is the default location of the release pipeline has been moved to
https://prod.ci.gpdb.pivotal.io/teams/main/pipelines/RELEASE:diskquota:gpdb
due to the requirement from the release team.

beeender:
Talked with @50wu  server/published/gpdb6/server-rc-(.*)-ubuntu18.04_x86_64.debug.tar.gz is built for every commit of the gpdb repo. So it is good for our use case.

concourse/fly.sh
```diff
@@ -121,6 +121,6 @@ fi
 echo ""
 echo "================================================================================"
 echo "Remeber to set the the webhook URL on GitHub:"
-echo "https://extensions.ci.gpdb.pivotal.io/api/v1/teams/main/pipelines/${pipeline_name}/resources/${hook_res}/check/webhook?webhook_token=<hook_token>"
+echo "https://<target>.ci.gpdb.pivotal.io/api/v1/teams/main/pipelines/${pipeline_name}/resources/${hook_res}/check/webhook?webhook_token=<hook_token>"
```
Sasasu, line 124: `fly targets | awk '{if ($1 == "${target}") {print $2}}'`
this will return the URL of current concourse server

yihong0618 (approved):
LGTM

Sasasu (approved):

PR #199, opened 2022-04-01
Fix tablespace per segment ratio

fanfuxiaoran:
If one tablespace segratio is configured,
this commit will make sure that ethier
the alredy existed (schema/role)_tablespace_quota
or new configs after this, will have the
same segratio.

1) Add a new quota type: TABLESPACE_QUOTA
used to store the tablespace segratio value
in quota_config table.
When set_per_segment_quota("xx","1.0) is
called, a new config will be added like:

      ` 97103 |         4 |            0 |        1`

   0 means invalid quota configed
   4 means TABLE_SPACE_QUOTA

2) Modify set_quota_config_internal to
support insert/delete/update new type config

3) Add function get_per_segment_ratio
it is called when insert new
(schema/role)_tablespace_quota.
"for share" is used to query segratio,it only
can keep the segratio is consistent in
"read committed" level.

PR #202, opened 2022-04-11
Fix un-expected remove when two tablespace-quota has a same role or schema

liuxueyang:
When there are two namespace_tablespace_quota configurations with the
same schema but different tablespaces, removing one quota will remove
another one automatically. The same bug happens for the
role_tabelspace_quota case. The root cause of this bug is that we
can't distinguish these two quotas using only quotaType and targetOid
of diskquota.quota_config.

This PR fixes this by changing the meaning of targetOid column of
diskquota.quota_config table. When quotaType is
role_quota/schema_quota, targetOid refers to role_oid/schema_oid. When
quotaType is role_tablespace_quota/namespace_tablespace_quota,
targetOid refers to the rowId column of diskquota.target. This will
make it possible to distinguish two namespace_tablespace_quota with
the same schema and different tablespace. Because their rowId will be
different.

For the new quota type TABLESPACE_QUOTA. It is not a real quota, we
only use it for configuration. So we won't insert quotas with this
type into quota_info.

### DIFF
--- a/diskquota--2.0--1.0.sql
+++ b/diskquota--2.0--1.0.sql
@@ -63,7 +63,7 @@ DROP FUNCTION diskquota.show_relation_cache_all_seg();
 -- UDF end
 
 -- table part
--- clean up schema_tablespace quota, rolsize_tablespace quota AND tablespace quota
+-- clean up NAMESPACE_TABLESPACE_QUOTA(2), ROLE_TABLESPACE_QUOTA(3) and TABLESPACE_QUOTA(4)
 DELETE FROM diskquota.quota_config WHERE quotatype in (2, 3, 4);
 
 DROP TABLE diskquota.target;
--- a/diskquota--2.0--1.0.sql
+++ b/diskquota--2.0--1.0.sql
@@ -63,8 +63,8 @@ DROP FUNCTION diskquota.show_relation_cache_all_seg();
 -- UDF end
 
 -- table part
--- clean up schema_tablespace quota AND rolsize_tablespace quota
-DELETE FROM diskquota.quota_config WHERE quotatype = 2 or quotatype = 3;
+-- clean up schema_tablespace quota, rolsize_tablespace quota AND tablespace quota
+DELETE FROM diskquota.quota_config WHERE quotatype in (2, 3, 4);
 
 DROP TABLE diskquota.target;
 
--- a/diskquota_utility.c
+++ b/diskquota_utility.c
@@ -92,7 +92,7 @@ static void        dq_object_access_hook(ObjectAccessType access, Oid classId, O
 static const char *ddl_err_code_to_err_message(MessageResult code);
 static int64       get_size_in_mb(char *str);
 static void set_quota_config_internal(Oid targetoid, int64 quota_limit_mb, QuotaType type, float4 segratio, Oid spcoid);
-static void set_target_internal(Oid primaryoid, Oid spcoid, int64 quota_limit_mb, QuotaType type);
+static int  set_target_internal(Oid primaryoid, Oid spcoid, int64 quota_limit_mb, QuotaType type);
 static float4 get_per_segment_ratio(Oid spcoid);
 static bool   to_delete_quota(QuotaType type, int64 quota_limit_mb, float4 segratio);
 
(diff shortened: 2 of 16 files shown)


diskquota_utility.c
```diff
@@ -795,8 +796,8 @@ set_role_tablespace_quota(PG_FUNCTION_ARGS)
 	}
 
 	SPI_connect();
-	set_target_internal(roleoid, spcoid, quota_limit_mb, ROLE_TABLESPACE_QUOTA);
-	set_quota_config_internal(roleoid, quota_limit_mb, ROLE_TABLESPACE_QUOTA, INVALID_SEGRATIO, spcoid);
+	row_id = set_target_internal(roleoid, spcoid, quota_limit_mb, ROLE_TABLESPACE_QUOTA);
+	set_quota_config_internal(row_id, quota_limit_mb, ROLE_TABLESPACE_QUOTA, INVALID_SEGRATIO, spcoid);
```
xuebinsu, line 800: Is this transactional?

quotamodel.c
```diff
@@ -1237,12 +1236,29 @@ do_load_quotas(void)
...
+	ret = SPI_execute_with_args(
 	        "SELECT c.targetOid, c.quotaType, c.quotalimitMB, COALESCE(c.segratio, 0) AS segratio, "
-	        "COALESCE(t.tablespaceoid, 0) AS tablespaceoid "
+	        "COALESCE(t.tablespaceoid, 0) AS tablespaceoid, COALESCE(t.primaryOid, 0) AS primaryoid "
 	        "FROM diskquota.quota_config AS c LEFT OUTER JOIN diskquota.target AS t "
-	        "ON c.targetOid = t.primaryOid and c.quotaType = t.quotaType",
-	        true, 0);
+	        "ON ((c.targetOid = t.primaryOid AND c.quotaType IN ($1, $2)) OR (c.targetOid = t.rowId "
+	        "AND c.quotaType IN ($3, $4))) AND c.quotaType = t.quotaType "
+	        "WHERE c.quotaType <> $5",
```
xuebinsu, line 1245: Why doing this filtering?

diskquota_utility.c
liuxueyang: Those two function calls are in the same transaction. They are wrapped in `SPI_connect` and `SPI_finish`.

diskquota--2.0--1.0.sql
```diff
@@ -63,8 +63,8 @@ DROP FUNCTION diskquota.show_relation_cache_all_seg();
 -- UDF end
 
 -- table part
--- clean up schema_tablespace quota AND rolsize_tablespace quota
-DELETE FROM diskquota.quota_config WHERE quotatype = 2 or quotatype = 3;
+-- clean up schema_tablespace quota, rolsize_tablespace quota AND tablespace quota
+DELETE FROM diskquota.quota_config WHERE quotatype in (2, 3, 4);
```
Sasasu, line 67: nice.
`3` is for #199
`4` is for this PR.

diskquota_utility.c
Sasasu: the `serial` type is not continuous.
it is (never reused) (always incremental), and (not continuous).

no problem here

Sasasu (approved):

diskquota_utility.c
fanfuxiaoran: > Those two function calls are in the same transaction. They are wrapped in `SPI_connect` and `SPI_finish`.

They are in a transaction,  not because the SPI, due to they are in one UDF "set_role_tablespace_quota",

diskquota_utility.c
```diff
@@ -994,40 +999,63 @@ set_target_internal(Oid primaryoid, Oid spcoid, int64 quota_limit_mb, QuotaType
...
+		                ObjectIdGetDatum(spcoid),
+		        },
+		        NULL, false, 0);
+		if (ret != SPI_OK_DELETE_RETURNING)
+			elog(ERROR, "cannot delete item from target setting table, error code %d", ret);
+
+		is_null = false;
+		v       = SPI_getbinval(SPI_tuptable->vals[0], SPI_tuptable->tupdesc, 1, &is_null);
+		Assert(is_null == false);
+		row_id = DatumGetInt32(v);
```
fanfuxiaoran, line 1054: How about return an invalid oid when delete from target table?

fanfuxiaoran (approved):

xuebinsu (approved):

PR #203, opened 2022-04-11
Reset memory account to fix memory leak

xuebinsu:
Diskquota workers dispatch queries periodically. For each dispatch,
one or more memory accounts will be created to keep track of the
memory usage. These accounts should be reset after the query
finishes to avoid memory leak.

Part of the stack trace of account creation looks like this:

```
   fun:CreateMemoryAccountImpl
   fun:MemoryAccounting_CreateAccount
   fun:serializeNode
   fun:cdbdisp_buildPlanQueryParms
   fun:cdbdisp_dispatchX
   fun:CdbDispatchPlan
   fun:standard_ExecutorStart
   fun:_SPI_pquery.constprop.11
   fun:_SPI_execute_plan
   fun:SPI_execute
   fun:do_load_quotas
   fun:load_quotas
   fun:refresh_disk_quota_model
   fun:disk_quota_worker_main
   fun:StartBackgroundWorker
   fun:do_start_bgworker
   fun:maybe_start_bgworker
```

This patch fixes the memory leak issue by calling the reset function 
at the end of the worker loop.

### DIFF
--- a/diskquota.c
+++ b/diskquota.c
@@ -422,7 +422,7 @@ disk_quota_worker_main(Datum main_arg)
 		/* Do the work */
 		if (!diskquota_is_paused()) refresh_disk_quota_model(false);
 
-		/* Reset memory account to fix memry leak */
+		/* Reset memory account to fix memory leak */
 		MemoryAccounting_Reset();
 		worker_increase_epoch(MyDatabaseId);
 	}
--- a/diskquota.c
+++ b/diskquota.c
@@ -422,6 +422,8 @@ disk_quota_worker_main(Datum main_arg)
 		/* Do the work */
 		if (!diskquota_is_paused()) refresh_disk_quota_model(false);
 
+		/* Reset memory account to fix memry leak */
+		MemoryAccounting_Reset();
 		worker_increase_epoch(MyDatabaseId);
 	}
 


liuxueyang (approved):

Sasasu:
please rewrite the commit message like this

```
Reset memory account to fix memory leak.

GPDB create a new memory account when serializing data and does not free them.

   fun:CreateMemoryAccountImpl
   fun:MemoryAccounting_CreateAccount
   fun:serializeNode
   fun:cdbdisp_buildPlanQueryParms
   fun:cdbdisp_dispatchX
   fun:CdbDispatchPlan
   fun:standard_ExecutorStart
   fun:_SPI_pquery.constprop.11
   fun:_SPI_execute_plan
   fun:SPI_execute
   fun:do_load_quotas
   fun:load_quotas
   fun:refresh_disk_quota_model
   fun:disk_quota_worker_main
   fun:StartBackgroundWorker
   fun:do_start_bgworker
   fun:maybe_start_bgworker

GPDB will reset all memory account after all transactions are finished in `PostgresMain`, only psql client will use the code path.
But if under SPI, it will be a memory leak.

fix the memory leak in diskquota's codebase is a workaround, by resetting all memory account at the end of the worker loop.
```

At least, include 1. where and how the leak happend. 2. why fix in diskquota's code (optional if only relate to diskquota's logic). 3. how the fix work? (optional if the fix is simple)

please don't repeat how diskquota works.

Sasasu (approved):

PR #89, opened 2021-11-25
Add support for dispatching blockmap to segment servers.

higuoxing:

PR #206, opened 2022-04-14
some fix for release CI

Sasasu:
release CI use naptime = 2.
add some `diskquota.wait_for_worker_new_epoch()` to make CI happy.

`invalidate_database_blackmap` on segment does work as expected, the old test is ﬂaky.
fix by correcting the search key. the key is different on master and segment.

### DIFF
--- a/quotamodel.c
+++ b/quotamodel.c
@@ -124,6 +124,7 @@ struct BlackMapEntry
 	Oid    tablespaceoid;
 	uint32 targettype;
 	/*
+	 * TODO refactor this data structure
 	 * QD index the blackmap by (targetoid, databaseoid, tablespaceoid, targettype).
 	 * QE index the blackmap by (relfilenode).
 	 */
@@ -1499,7 +1500,7 @@ invalidate_database_blackmap(Oid dbid)
 	hash_seq_init(&iter, disk_quota_black_map);
 	while ((entry = hash_seq_search(&iter)) != NULL)
 	{
-		if (entry->databaseoid == dbid)
+		if (entry->databaseoid == dbid || entry->relfilenode.dbNode == dbid)
 		{
 			hash_search(disk_quota_black_map, entry, HASH_REMOVE, NULL);
 		}
--- a/tests/regress/expected/test_activetable_limit.out
+++ b/tests/regress/expected/test_activetable_limit.out
@@ -20,14 +20,34 @@ SELECT diskquota.set_schema_quota('s', '1 MB');
  
 (1 row)
 
-CREATE TABLE s.t1(i int) DISTRIBUTED BY (i);
-INSERT INTO s.t1 SELECT generate_series(1,100000); -- expect failed. diskquota should works. activetable = 1
-CREATE TABLE s.t2(i int) DISTRIBUTED BY (i);
-INSERT INTO s.t2 SELECT generate_series(1,100000);
+SELECT diskquota.wait_for_worker_new_epoch();
+ wait_for_worker_new_epoch 
+---------------------------
+ t
+(1 row)
+
+CREATE TABLE s.t1(i int) DISTRIBUTED BY (i); -- activetable = 1
+INSERT INTO s.t1 SELECT generate_series(1, 100000); -- ok. diskquota soft limit dose not check when first write
+SELECT diskquota.wait_for_worker_new_epoch();
+ wait_for_worker_new_epoch 
+---------------------------
+ t
+(1 row)
+
+CREATE TABLE s.t2(i int) DISTRIBUTED BY (i); -- activetable = 2
+INSERT INTO s.t2 SELECT generate_series(1, 10);  -- expect failed
 ERROR:  schema's disk space quota exceeded with name:s
-CREATE TABLE s.t3(i int) DISTRIBUTED BY (i);
-INSERT INTO s.t3 SELECT generate_series(1,100000); -- should not crash. activetable = 3
+CREATE TABLE s.t3(i int) DISTRIBUTED BY (i); -- activetable = 3 should not crash.
+INSERT INTO s.t3 SELECT generate_series(1, 10);  -- expect failed
 ERROR:  schema's disk space quota exceeded with name:s
+-- Q: why diskquota still works when activetable = 3?
+-- A: the activetable limit by shmem size, calculate by hash_estimate_size()
+--    the result will bigger than sizeof(DiskQuotaActiveTableEntry) * max_active_tables
+--    the real capacity of this data structure based on the hash conflict probability.
+--    so we can not predict when the data structure will be fill in fully.
+--
+--    this test case is use less, remove this if anyone dislike it.
+--    but the hash capacity is smaller than 6, so the test case works for issue 51
 DROP EXTENSION diskquota;
 -- wait worker exit
 \! sleep 1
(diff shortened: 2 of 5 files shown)


Sasasu:
@beeender I think on your machine the clean blackmap is flaky because some pgcli's default behavior.
pgcli will stop executing when some error occurred.
```
INSERT INTO s.t3 SELECT generate_series(1, 10000); -- expect failed
DROP EXTENSION diskquota;
```
paste this into pgcli. and run, if the insert errors, the drop will not be executed.

fix this by editing `.config/pgcli/config` add `on_error = RESUME`

tests/regress/expected/test_clean_blackmap_after_drop.out
```diff
@@ -19,8 +19,8 @@ SELECT diskquota.wait_for_worker_new_epoch();
  t
 (1 row)
 
-INSERT INTO b SELECT generate_series(1, 100000); -- fail
-ERROR:  role's disk space quota exceeded with name:40716  (seg2 127.0.0.1:6004 pid=1245042)
+INSERT INTO b SELECT generate_series(1, 100000000); -- fail
```
Sasasu, line 22: for all hardlimit test. `sizeof(generate_series(1, 100000000))` should be bigger than `naptime*write speed (60MiB/s on my machine) = 120MiB`

tests/regress/sql/test_activetable_limit.sql
```diff
@@ -22,13 +22,28 @@ CREATE EXTENSION diskquota;
...
+INSERT INTO s.t2 SELECT generate_series(1, 10000); -- expect failed
+INSERT INTO s.t3 SELECT generate_series(1, 10000); -- expect failed
+-- Q: why diskquota still works when activetable = 3?
+-- A: the activetable limit by shmem size, calculate by hash_estimate_size()
+--    the result will bigger than sizeof(DiskQuotaActiveTableEntry) * max_active_tables
+--    the real capacity of this data structure based on the hash conflict probability.
+--    so we can not predict when the data structure will be fill in fully.
+--
+--    this test case is use less, remove this if anyone dislike it.
+--    but the hash capacity is smaller than 6, so the test case works for issue 51
```
Sasasu, line 46: #51

beeender:
> I think on your machine the clean blackmap is flaky because some pgcli's default behavior.
> pgcli will stop executing when some error occurred.

I don't think I was using pgcli at that time :(

Sasasu:
> > I think on your machine the clean blackmap is flaky because some pgcli's default behavior.
> > pgcli will stop executing when some error occurred.
> 
> I don't think I was using pgcli at that time :(

Ok, it must be this
```diff
-if (entry->databaseoid == dbid)
+if (entry->databaseoid == dbid || entry->relfilenode.dbNode == dbid)
```

quotamodel.c
```diff
@@ -1499,7 +1500,7 @@ invalidate_database_blackmap(Oid dbid)
 	hash_seq_init(&iter, disk_quota_black_map);
 	while ((entry = hash_seq_search(&iter)) != NULL)
 	{
-		if (entry->databaseoid == dbid)
+		if (entry->databaseoid == dbid || entry->relfilenode.dbNode == dbid)
```
xuebinsu, line 1503: Why do they differ?

Sasasu: I did not know. you did this.
https://github.com/greenplum-db/diskquota/pull/58 https://github.com/greenplum-db/diskquota/pull/89

xuebinsu: I mean,  why would `entry->databaseoid != entry->relfilenode.dbNode`?

Sasasu: you add the logic.
on master use entry->databaseoid. on segment use entry->relfilenode.dbNode

Sasasu:
```
/* global blacklist for which exceed their quota limit */
struct BlackMapEntry
{
	Oid		targetoid;
	Oid		databaseoid;
	Oid 		tablespaceoid;
	uint32		targettype;
	/*
	 * QD index the blackmap by (targetoid, databaseoid, tablespaceoid, targettype).
	 * QE index the blackmap by (relfilenode).
	 */
	RelFileNode		relfilenode;
};
```
and there even has a comments for it

xuebinsu: Actually, I dont know either. I think they always have the same value.

tests/regress/sql/test_activetable_limit.sql
```diff
@@ -22,13 +22,26 @@ CREATE EXTENSION diskquota;
...
+CREATE TABLE s.t3(i int) DISTRIBUTED BY (i); -- activetable = 3 should not crash.
+INSERT INTO s.t3 SELECT generate_series(1, 10);  -- expect failed
+
+-- Q: why diskquota still works when activetable = 3?
+-- A: the activetable limit by shmem size, calculate by hash_estimate_size()
+--    the result will bigger than sizeof(DiskQuotaActiveTableEntry) * max_active_tables
+--    the real capacity of this data structure based on the hash conflict probability.
+--    so we can not predict when the data structure will be fill in fully.
+--
+--    this test case is use less, remove this if anyone dislike it.
```
yihong0618, line 43: use less -> useless ^ ^

yihong0618 (approved):

tests/regress/expected/test_activetable_limit.out
```diff
@@ -20,14 +20,34 @@ SELECT diskquota.set_schema_quota('s', '1 MB');
  
 (1 row)
 
-CREATE TABLE s.t1(i int) DISTRIBUTED BY (i);
-INSERT INTO s.t1 SELECT generate_series(1,100000); -- expect failed. diskquota should works. activetable = 1
```
beeender, line 24: How does it pass before? The only logic changing is the newly added `wait_for_worker_new_epoch`. Wouldn't it be flaky still?

Sasasu: diskquota soft limit does not check when the first write.
the behavior is correct. expect is wrong

beeender (approved):

quotamodel.c
beeender: Interesting, don't merge for now, I am checking

beeender: OK, it took me quite some time to understand this, this change looks OK to me.
I have some basic refactor ideas, we have been falling into the same trap a few times, let's fix this.
@higuoxing would you please help to refactor this when you are free? Let's discuss first, I can do the refactor if you are not available.

Ship this :shipit:

higuoxing: The original idea is:

Soft-limit on tables/indexes is performed by query dispatcher (QD) and it relies on the `object_access_hook` function. QD checks whether the relation should be qutaed by comparing the relation oid, tablespace oid and database oid with the records in the blackmap/rejectmap. That's why QD searches the blackmap/rejectmap entry by (targetoid (relation oid), databaseoid, tablespaceoid, targettype).

Hard-limit relies on the `smgrextend` hook and the hook is only called on segment servers. The query executors check whether the relatoin should be quotaed during its extending by compare the relation's underlying `relfilenode` with the records in the blackmap/rejectmap. That's why QE searches the blackmap/rejectmap entry by (relfilenode).

higuoxing: In theory, they represent the same thing. But for convenience (maybe it's not convenient and causing confusion), the rejectmap entry's databaseoid (`entry->databaseoid`) is zero and its real database oid is in `entry->relfilenode.dbNode`. https://github.com/greenplum-db/diskquota/blob/37a4caf6474c708aca5aeb5475212eb83fd3b782/quotamodel.c#L1779

PR #208, opened 2022-04-18
Fix set quota with upper case object name

Sasasu:
the code transform all object name to lower case on purpose.

    create schema "S1";
    select * from diskquota.set_schema_quota("S1", '1MB');
    ERROR:  schema "s1" does not exist

### DIFF
--- a/diskquota_utility.c
+++ b/diskquota_utility.c
@@ -690,6 +690,31 @@ ddl_err_code_to_err_message(MessageResult code)
 	}
 }
 
+static Datum
+__get_oid_auto_case_convert(Oid (*f)(const char *name, bool missing_ok), const char *name)
+{
+	char *b   = NULL;
+	int   l   = strlen(name);
+	Oid   ret = InvalidOid;
+
+	if (l > 2 && name[0] == '"' && name[l - 1] == '"')
+	{
+		// object name wrapped by '"'. eg: "foo"
+		b = palloc(l);
+		StrNCpy(b, name + 1, l - 1); // trim the '"'. unlike strncpy, StrNCpy will ensure b[l-1] = '\0'
+	}
+	else
+	{
+		// lower the object name if not wrapped by '"'
+		b = str_tolower(name, strlen(name), DEFAULT_COLLATION_OID);
+	}
+
+	ret = f(b, false);
+
+	pfree(b);
+	return ret;
+}
+
 /*
  * Set disk quota limit for role.
  */
@@ -707,8 +732,7 @@ set_role_quota(PG_FUNCTION_ARGS)
 	}
 
 	rolname = text_to_cstring(PG_GETARG_TEXT_PP(0));
-	rolname = str_tolower(rolname, strlen(rolname), DEFAULT_COLLATION_OID);
-	roleoid = get_role_oid(rolname, false);
+	roleoid = __get_oid_auto_case_convert(get_role_oid, rolname);
 	check_role(roleoid, rolname);
 
 	sizestr        = text_to_cstring(PG_GETARG_TEXT_PP(1));
@@ -742,8 +766,7 @@ set_schema_quota(PG_FUNCTION_ARGS)
 	}
 
 	nspname      = text_to_cstring(PG_GETARG_TEXT_PP(0));
-	nspname      = str_tolower(nspname, strlen(nspname), DEFAULT_COLLATION_OID);
-	namespaceoid = get_namespace_oid(nspname, false);
+	namespaceoid = __get_oid_auto_case_convert(get_namespace_oid, nspname);
 
 	sizestr        = text_to_cstring(PG_GETARG_TEXT_PP(1));
 	sizestr        = str_tolower(sizestr, strlen(sizestr), DEFAULT_COLLATION_OID);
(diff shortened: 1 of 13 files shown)


Sasasu:
```
CREATE SCHEMA SX;
CREATE TABLE SX.a(i int) DISTRIBUTED BY (i);
SELECT diskquota.set_schema_quota('SX', '1MB');
ERROR:  schema "SX" does not exist
```

ok... this is why transform all object name to lower case on purpose.

Sasasu:
modify to check lowercase first. then chack the original input.

diskquota_utility.c
```diff
@@ -704,9 +704,14 @@ set_role_quota(PG_FUNCTION_ARGS)
 		ereport(ERROR, (errcode(ERRCODE_INSUFFICIENT_PRIVILEGE), errmsg("must be superuser to set disk quota limit")));
 	}
 
-	rolname = text_to_cstring(PG_GETARG_TEXT_PP(0));
-	rolname = str_tolower(rolname, strlen(rolname), DEFAULT_COLLATION_OID);
-	roleoid = get_role_oid(rolname, false);
+	rolname      = text_to_cstring(PG_GETARG_TEXT_PP(0));
```
beeender, line 707: instead of using `text` as the role/schema/table name, can we just use `regclass` like `pg_table_size()`?

`pg_table_size()` is defined as:

```
pg_table_size ( regclass ) → bigint
```

PG will automatically convert the name to the oid. getting the relation oid is as simple as
 
```
Datum
pg_table_size(PG_FUNCTION_ARGS)
{
	Oid			relOid = PG_GETARG_OID(0);
    ...
}
```

Plus this handles all naming rules automatically:

```

isolation2test=# \dt+
                        List of relations
 Schema | Name | Type  | Owner | Storage |  Size   | Description
--------+------+-------+-------+---------+---------+-------------
 public | T1   | table | cc    | heap    | 32 kB   |
 public | t1   | table | cc    | heap    | 0 bytes |
 public | t 1  | table | cc    | heap
(2 rows)

isolation2test=# select pg_table_size('"t1"');
 pg_table_size
---------------
             0
(1 row)

isolation2test=# select pg_table_size('"T1"');
 pg_table_size
---------------
         32768
(1 row)

isolation2test=# select pg_table_size('"Tt1"');
ERROR:  relation "Tt1" does not exist
LINE 1: select pg_table_size('"Tt1"');
isolation2test=# select pg_table_size('"t 1"');
 pg_table_size
---------------
             0
(1 row)
```

Will this make our code simpler?

xuebinsu: Unfortunately, PG does not provide something like "regrole", "regnamespace" and "regtablespace". 

But we can implement the same behavior here by ourselves if desired.

Sasasu: I add a `__get_oid_case_insensitive()`, usage: `roleoid = __get_oid_case_insensitive(get_role_oid, rolname);`

is simpler than before.

diskquota_utility.c
```diff
@@ -688,6 +688,23 @@ ddl_err_code_to_err_message(MessageResult code)
...
 }
 
+static Datum
+__get_oid_case_insensitive(Oid (*f)(const char *name, bool missing_ok), const char *name)
+{
+	char *lower       = str_tolower(name, strlen(name), DEFAULT_COLLATION_OID);
+	Oid   originalOid = f(name, true);
+	Oid   lowerOid    = f(lower, true);
+
+	if (!OidIsValid(lowerOid) && !OidIsValid(originalOid))
```
beeender, line 698: I would like to avoid the uncertainty of function. This cannot handle `Role1` and `role1` exist at the same time.
How about we follow the PG rules in general and make it a bit simpler?
```
if (strlen(name) > 2 && name[0] == '"' && name[strlen(name) - 1] == '"')
{
// always use the original
}
else
{
// always convert to lower
}
```

xuebinsu:
Maybe we can use `SplitIdentifierString()`

Sasasu:
> Maybe we can use SplitIdentifierString()

`SplitIdentifierString(char *rawstring, char separator, List **namelist)`

`SplitIdentifierString` needs alloc a list<char*>, which is very very very complex.

tests/regress/expected/test_tablespace_schema_perseg.out
```diff
@@ -268,15 +268,15 @@ SELECT distinct(segratio) FROM diskquota.quota_config, pg_namespace, diskquota.t
         3
 (1 row)
 
-SELECT tablespace_name, per_seg_quota_ratio FROM diskquota.show_segment_ratio_quota_view where tablespace_name in ('schemaspc_perseg2', 'schemaspc_perseg');
+SELECT tablespace_name, per_seg_quota_ratio FROM diskquota.show_segment_ratio_quota_view where tablespace_name in ('Schemaspc_perseg2', 'schemaspc_perseg');
```
beeender, line 271: `'"Schemaspc_perseg2"'`

beeender: ignore me, wrong comment

diskquota_utility.c
```diff
@@ -688,6 +688,31 @@ ddl_err_code_to_err_message(MessageResult code)
...
+{
+	char *b   = NULL;
+	int   l   = strlen(name);
+	Oid   ret = InvalidOid;
+
+	if (l > 2 && name[0] == '"' && name[l - 1] == '"')
+	{
+		// object name wrapped by '"'. eg: "foo"
+		b = palloc(l);
+		StrNCpy(b, name + 1, l - 1); // trim the '"'. unlike strncpy StrNCpy will ensure b[l-1] = '\0'
```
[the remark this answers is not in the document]
beeender: hmmm, ok, it is `StrNCpy` magic

beeender (approved):

xuebinsu (approved):

PR #209, opened 2022-04-18
Reformat quota exceeded message

beeender:


Sasasu (approved):
also need to update 'test 1.0 quota in 2.0 binary'
no need to update 'test 2.0 quota in 1.0 binary'

they are different test result files.

liuxueyang (approved):

PR #211, opened 2022-04-19
SPI & entry clear in refresh_blackmap

beeender:
- refresh_blackmap doesn't need SPI calls.
- Clearing should happen within the same lock of refreshing entries.
  Since there will be a short time that other processes see a empty
  blackmap while checking the hard limit.


quotamodel.c
```diff
@@ -1860,6 +1847,12 @@ refresh_blackmap(PG_FUNCTION_ARGS)
 
 	/* Flush the content of local_blackmap to the global blackmap. */
 	LWLockAcquire(diskquota_locks.black_map_lock, LW_EXCLUSIVE);
+
+	/* Clear blackmap entries. */
+	hash_seq_init(&hash_seq, disk_quota_black_map);
+	while ((blackmapentry = hash_seq_search(&hash_seq)) != NULL)
+		hash_search(disk_quota_black_map, &blackmapentry->keyitem, HASH_REMOVE, NULL);
+
 	hash_seq_init(&hash_seq, local_blackmap);
```
Sasasu, line 1856: https://github.com/greenplum-db/diskquota/pull/138#discussion_r807504101

local_blackmap can be a simple sorted array and not a global variable. I want to put this thing aside to do the release.
but we still fixing some rare bugs, this is not the right way.

this PR is OK to me, but this part code needs a big but simple refactor after we push a release.

Sasasu (approved):

quotamodel.c
beeender: > but we still fixing some rare bugs, this is not the right way.

yeah, but this one costs no effort. Just saw it and fix it.

higuoxing (approved):
Good catch! LGTM with one nit.

beeender:
CI fails, not sure if it is related. Will check before merge

Sasasu (approved):

Sasasu:
I made a branch to resolve the conflicts. https://github.com/greenplum-db/diskquota/commit/812747ec8d01d107b3f02cbba75d15eb834bfbe3

PR #212, opened 2022-04-20
Minimize the risk of stale relfilenode

xuebinsu:
- Check for invalidation messages each time before SearchSysCache().
- Do not remove a relation if the relfilenode is stale.


xuebinsu:
Closed in favor of #217 .

PR #215, opened 2022-04-21
Reject set quota for superuser

fanfuxiaoran:


tests/regress/expected/test_role.out
```diff
@@ -102,6 +102,16 @@ SELECT diskquota.wait_for_worker_new_epoch();
 INSERT INTO b SELECT generate_series(1,100);
 -- expect insert succeed
 INSERT INTO b2 SELECT generate_series(1,100);
+-- superuser is blocked to set quota
+SELECT rolname from pg_roles where rolsuper=true;
+ rolname 
+---------
+ xiwang
```
beeender, line 109: this will fail CI :grin:

fanfuxiaoran: haha

beeender: Try this trick?

```
SELECT concat("sys_role_", rolename) from pg_roles where rolsuper=true;
```

And ignore the string begins with `sys_role_`

tests/regress/expected/test_role.out
```diff
@@ -102,6 +102,16 @@ SELECT diskquota.wait_for_worker_new_epoch();
...
+(1 row)
+
+\gset
+select diskquota.set_role_quota(:'rolname', '1mb');
```
beeender, line 113: uh, you have a trick! then I think just `--start-ignore` the above query result could just work?

fanfuxiaoran: Yeah, this works. Have fixed!

Sasasu (approved):
add this to `init_file` to make our CI happy

```
-- start_matchsubs
m/^ERROR:  Can not set disk quota from system owner: \c+/
s/^ERROR:  Can not set disk quota from system owner: \c+/ERROR:  Can not set disk quota from system owner:/
-- end_matchsubs
```

beeender (approved):

liuxueyang (approved):

radarwave:
As we only limit the system owner, other super users are not affected. Maybe we should change the commit message title to 'system owner' to avoid future confusion.

PR #216, opened 2022-04-25
Hint message when launcher cannot be connected

beeender:


Sasasu:
https://github.com/greenplum-db/diskquota/blob/7d4d66d9b2efa3ac785d3d7fbb5547a69fbc08c2/diskquota_utility.c#L358-L359
I think need to add a hint message here.

btw, I think the error code needs clean up. at least, add ERR_WORKER_CRASH and ERR_LAUNCHER_CRASH and remove ERR_PENDING ERR_EXCEED

beeender:
> https://github.com/greenplum-db/diskquota/blob/7d4d66d9b2efa3ac785d3d7fbb5547a69fbc08c2/diskquota_utility.c#L358-L359
> 
> I think need to add a hint message here.

Can do, will add a hint there as well.

> btw, I think the error code needs clean up. at least, add ERR_WORKER_CRASH and ERR_LAUNCHER_CRASH and remove ERR_PENDING ERR_EXCEED

`extension_ddl_message->result` is supposed to have an initial value `ERR_PENDING` and all other values are set in the launcher process. AFAIK, `extension_ddl_message` is designed for client -> launcher process communication (one way) for now. And the result is like launcher set a response code and send it back to the caller -- the client.

`ERR_LAUNCHER_CRASH` is not proper to be set at any place I think? If it is added before:
```
			ereportif(kill(launcher_pid, 0) == -1 && errno == ESRCH, // do existence check
			          ERROR, (errmsg("[diskquota] diskquota launcher pid = %d no longer exists", launcher_pid)));

```
Then no one will check the value anymore, since an error will be emitted after that.

`ERR_WORKER_CRASH` same, it is not a response from the launcher process.

`ERR_PENDING` as I mentioned, it is the initial value, any reason we should remove it?

`ERR_EXCEED` why remove this one? It means "the number of database exceeds the maximum" which is an error responce while creating the extension. Anything I missed here?

Sasasu:
> ERR_PENDING as I mentioned, it is the initial value, any reason we should remove it?

too general. no idea about is the actual error.

> ERR_EXCEED why remove this one?

too little context.

> extension_ddl_message is designed for client -> launcher process communication (one way) for now.

the current IPC infrastructure is 2-way IPC. the caller can send a message to launcher, and launcher can send an error code back.
a little more modification can make the error message more friendly. like some `error_buf[128]` in shmem.

but this is refactor thing, we can do it at any time. not in this PR's scope.
---

for this PR. I think the error message does not take effect.

timeline:
    - gpdb boot
    - launcher boot, no diskquota database exit
    - create extension.
    - check if launcher exist. not. error out at [diskquota_utility.c#L358-L359](https://github.com/greenplum-db/diskquota/blob/7d4d66d9b2efa3ac785d3d7fbb5547a69fbc08c2/diskquota_utility.c#L358-L359)

xuebinsu:
How about checking whether database "diskquota" exists when `CREATE EXTENSION` using the `object_access_hook()`?

beeender:
> for this PR. I think the error message does not take effect.

It does take effect:

```
	rc                             = kill(launcher_pid, SIGUSR1);
	LWLockRelease(diskquota_locks.extension_ddl_message_lock);
	if (rc == 0)
	{
		int count = WAIT_TIME_COUNT;

		while (count-- > 0)
		{
			CHECK_FOR_INTERRUPTS();
			rc = WaitLatch(&MyProc->procLatch, WL_LATCH_SET | WL_TIMEOUT | WL_POSTMASTER_DEATH, 100L);
			if (rc & WL_POSTMASTER_DEATH) break;

			ereportif(kill(launcher_pid, 0) == -1 && errno == ESRCH, // do existence check
			          ERROR,
					  (errmsg("[diskquota] diskquota launcher pid = %d no longer exists", launcher_pid),
					   errhint("The diskquota launcher process has been terminated for some reasons. Consider to restart the cluster to start it.")));

```

There is an `rc == 0` check before entering the block.

beeender:
> How about checking whether database "diskquota" exists when CREATE EXTENSION using the object_access_hook()?

Can do, but wouldn't that be too complex? I can try if you insist.

Sasasu (approved):
ok, rc = 0 check is a little complex.

better to remove the `kill(launcher_pid, 0) == -1 && errno == ESRCH` check.

beeender:
> better to remove the kill(launcher_pid, 0) == -1 && errno == ESRCH check.

different purpose, the later check if for the launcher process gets terminated after `SIGUSR1` sent.

Sasasu:
> different purpose, the latter check is for the launcher process gets terminated after SIGUSR1 sent.

both are fine.

xuebinsu:
> wouldn't that be too complex?

I don't think it will be complex. One `SPI_execute()` will do the job.

There are many reasons that can cause the "no-response" error. I suspect that it is still possible even when the diskquota launcher is alive (Probably due to some bugs, e.g., the launcher is waiting for a lock indefinitely).

But for the case when database "diskquota" does not exists, we can check that and give a clear message what should the user do. I think we should do that.

beeender:
> I don't think it will be complex. One `SPI_execute()` will do the job.
> 
> There are many reasons that can cause the "no-response" error. I suspect that it is still possible even when the diskquota launcher is alive (Probably due to some bugs, e.g., the launcher is waiting for a lock indefinitely).
> 
> But for the case that database "diskquota" does not exists, we can check that and give a clear message what should the user do. I think we should do that.
> 
> EDIT: we can also do this check in `diskquota_start_worker()`.

I got some other ideas. will change this.

Sasasu:
> I don't think it will be complex. One SPI_execute() will do the job.

'diskquota' database is really a  bad design, I think we should not add more code to ensure this logic.

if you want to ensure diskquota database always exists. you can boot a new process after database is ready, call `BackgroundWorkerInitializeConnection(NULL, NULL)` create the 'diskquota' database then boot the launcher.

fix bug point by point will case endless bug fixing. need do refactor to break the loop.

for now, I think adding some hint to ERR_PENDING is good enough.

Sasasu (approved):

beeender:
Some demonstration of the idea

https://github.com/greenplum-db/diskquota/tree/mc/ddl_message_refactor

To fully implement the idea, still need some time to move some code around.

@xuebinsu Can we merge this PR as it is and I will finish the refactor idea later.

xuebinsu (approved):

PR #217, opened 2022-04-25
Minimize the risk of stale relfilenodes

xuebinsu:
Since we do NOT take any lock when accessing info of a relation,
cache invalidation messages may not be delievered in time each time
we SeearchSysCache() for relfilenode. As a result, the relfilenode may
be stale and the table size may be incorrect.

This patch fixes this issue by

- Checking for invalidation messages each time before
SearchSysCache(RELOID).
- Not removing a relation from active table map if the relfilenode
is stale.

Besides, since we do not remove the relfilenode if it maps to no
valid relation oid, we need remove those dangling relfilenodes
when unlink() to avoid memory leak.

### DIFF
--- a/diskquota.c
+++ b/diskquota.c
@@ -396,16 +396,6 @@ disk_quota_worker_main(Datum main_arg)
 		CHECK_FOR_INTERRUPTS();
 
 		/*
-		 * Allow sinval catchup interrupts while sleeping
-		 *
-		 * This follows what the autovacuum launcher does
-		 *
-		 * TODO: Currently, this does not take effect because there is no
-		 * handler for it in disk_quota_sigusr1().
-		 */
-		EnableCatchupInterrupt();
-
-		/*
 		 * Background workers mustn't call usleep() or any direct equivalent:
 		 * instead, they may wait on their process latch, which sleeps as
 		 * necessary, but is awakened if postmaster dies.  That way the
@@ -413,8 +403,6 @@ disk_quota_worker_main(Datum main_arg)
 		 */
 		rc = WaitLatch(&MyProc->procLatch, WL_LATCH_SET | WL_TIMEOUT | WL_POSTMASTER_DEATH, diskquota_naptime * 1000L);
 		ResetLatch(&MyProc->procLatch);
-
-		DisableCatchupInterrupt();
 
 		// be nice to scheduler when naptime == 0 and diskquota_is_paused() == true
 		if (!diskquota_naptime) usleep(1);
--- a/diskquota.c
+++ b/diskquota.c
@@ -28,7 +28,6 @@
 #include "port/atomics.h"
 #include "storage/ipc.h"
 #include "storage/proc.h"
-#include "storage/sinval.h"
 #include "tcop/idle_resource_cleaner.h"
 #include "tcop/utility.h"
 #include "utils/builtins.h"
--- a/diskquota.c
+++ b/diskquota.c
@@ -398,7 +398,10 @@ disk_quota_worker_main(Datum main_arg)
 		/*
 		 * Allow sinval catchup interrupts while sleeping
 		 *
-		 * This follows what the Autovacuum launcher does
+		 * This follows what the autovacuum launcher does
+		 *
+		 * TODO: Currently, this does not take effect because there is no
+		 * handler for it in disk_quota_sigusr1().
 		 */
 		EnableCatchupInterrupt();
 
--- a/diskquota.c
+++ b/diskquota.c
@@ -28,6 +28,7 @@
 #include "port/atomics.h"
 #include "storage/ipc.h"
 #include "storage/proc.h"
+#include "storage/sinval.h"
 #include "tcop/idle_resource_cleaner.h"
 #include "tcop/utility.h"
 #include "utils/builtins.h"
@@ -395,6 +396,13 @@ disk_quota_worker_main(Datum main_arg)
 		CHECK_FOR_INTERRUPTS();
 
 		/*
+		 * Allow sinval catchup interrupts while sleeping
+		 *
+		 * This follows what the Autovacuum launcher does
+		 */
+		EnableCatchupInterrupt();
+
+		/*
 		 * Background workers mustn't call usleep() or any direct equivalent:
 		 * instead, they may wait on their process latch, which sleeps as
 		 * necessary, but is awakened if postmaster dies.  That way the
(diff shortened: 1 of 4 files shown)


Sasasu (approved):

gp_activetable.c
```diff
@@ -333,6 +358,9 @@ gp_fetch_active_tables(bool is_init)
 		local_active_table_oid_maps = pull_active_list_from_seg();
 		active_oid_list             = convert_map_to_string(local_active_table_oid_maps);
 
+		ereport(WARNING,
```
Sasasu, line 361: remember change this to DEBUG level

Sasasu (changes requested):

Sasasu (approved):
the double negative in `is_relfilenode_stale()` is not clear, but I think is fine.

diskquota.c
```diff
@@ -394,6 +395,13 @@ disk_quota_worker_main(Datum main_arg)
 
 		CHECK_FOR_INTERRUPTS();
 
+		/*
+		 * Allow sinval catchup interrupts while sleeping
+		 *
+		 * This follows what the Autovacuum launcher does
+		 */
+		EnableCatchupInterrupt();
```
beeender, line 403:
```
void
procsignal_sigusr1_handler(SIGNAL_ARGS)
{
	int			save_errno = errno;

	if (CheckProcSignal(PROCSIG_CATCHUP_INTERRUPT))
		HandleCatchupInterrupt();
```

The catchup interrupt is triggered by `SIGUSR1`

```
void
EnableCatchupInterrupt(void)
{
	for (;;)
	{
		catchupInterruptEnabled = 1;
		if (!catchupInterruptOccurred)
			break;
		catchupInterruptEnabled = 0;
		if (catchupInterruptOccurred)
			ProcessCatchupEvent();
	}

}
```

`EnableCatchupInterrupt` only does the real work (`ProcessCatchupEvent()') when `catchupInterruptOccurred` is set. However, the only place set it is in `HandleCatchupInterrupt()` which is triggered by `SIGUSR1`.

The problem is, we overwrite the `SIGUSR1` handler by `	pqsignal(SIGUSR1, disk_quota_sigusr1);` at the beginning of `xxx_main` function. This means, for diskquota laucher & bgworkers, the `HandleCatchupInterrupt()` has never been called. (unlike Autovacuum, see https://github.com/greenplum-db/gpdb/blob/6X_STABLE/src/backend/postmaster/autovacuum.c#L467 )

So, my doubt is, the `EnableCatchupInterrupt()` and `DisableCatchupInterrupt()` here DOES NOT really do anything.

I could be wrong, let me know if there are more details which I am no aware of.

xuebinsu: Good point! Now I know why the worker never receive the catchup interrupt.

Any idea how we should fix this?

beeender: We don't quite need the update-to-date syscache on bgworker I think? This file change won't be needed to fix the problem.

beeender (commented):
Looks good!

gp_activetable.c
```diff
@@ -599,6 +627,37 @@ is_relation_being_altered(Oid relid)
...
+ * NOTE: It will return false if the relation is currently uncommitted.
+ */
+static bool
+is_relfilenode_stale(Oid relOid, RelFileNode rnode)
+{
+	/*
+	 * Since we don't take any lock on relation, need to check for cache
+	 * invalidation messages manually.
+	 */
+	AcceptInvalidationMessages();
```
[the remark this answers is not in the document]
xuebinsu: I think so.

gp_activetable.c
```diff
@@ -164,6 +167,11 @@ active_table_hook_smgrunlink(RelFileNodeBackend rnode)
 {
 	if (prev_file_unlink_hook) (*prev_file_unlink_hook)(rnode);
 
+	/*
+	 * Since we do not remove the relfilenode if it maps to no valid
```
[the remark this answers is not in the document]
xuebinsu: The description here is not quite correct. I will update it.

quotamodel.c
```diff
@@ -1258,8 +1258,8 @@ do_load_quotas(void)
 		                errmsg("[diskquota] load_quotas SPI_execute failed: error code %d", ret)));
 
 	tupdesc = SPI_tuptable->tupdesc;
-	if (tupdesc->natts != NUM_QUOTA_CONFIG_ATTRS || ((tupdesc)->attrs[0])->atttypid != OIDOID ||
-	    ((tupdesc)->attrs[1])->atttypid != INT4OID || ((tupdesc)->attrs[2])->atttypid != INT8OID)
+	if (tupdesc && (tupdesc->natts != NUM_QUOTA_CONFIG_ATTRS || ((tupdesc)->attrs[0])->atttypid != OIDOID ||
```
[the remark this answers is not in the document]
xuebinsu: Revered.

gp_activetable.c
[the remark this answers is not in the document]
xuebinsu: Updated.

diskquota.c
xuebinsu: I think we need this handler. Maybe we can add this later. Added a TODO item.

fanfuxiaoran: As we have called "AcceptInvalidationMessages" where we need to fetch the latest catalog info, there is no need to call the  EnableCatchupInterrupt to catch catchup event.

gp_activetable.c
```diff
@@ -691,6 +719,25 @@ get_active_tables_oid(void)
 
 		if (relOid != InvalidOid)
```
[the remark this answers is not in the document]
xuebinsu: Tried that. This is not enough to reduce the error probability. We'd better do this for each `SearchSysCache()`.

diskquota.c
beeender: code without doing anything will be confusing. Plus `CheckProcSignal` is a private function, we cannot handle `PROCSIG_CATCHUP_INTERRUPT` only by design. We may need to investigate if all other signals are interested to us as well.

I kind of agree with Xiaoran that `AcceptInvalidationMessages` already does the job, we don't have to catch the signals anymore.

Information like the comments here should exist in the comment to describe the alternative way of handling the same issue. To keep the code short and clean.

xuebinsu: OK. I will revert it.

diskquota.c
```diff
@@ -28,6 +28,7 @@
 #include "port/atomics.h"
 #include "storage/ipc.h"
 #include "storage/proc.h"
+#include "storage/sinval.h"
```
beeender, line 31: remove this

beeender (commented):
LGTM, Let's merge if @fanfuxiaoran is fine with this as well.

fanfuxiaoran (approved):

diskquota.c
fanfuxiaoran: >EnableCatchupInterrupt()
The main thing it does is calling "ProcessCatchupEvent " to process catchup event and set  "catchupInterruptEnabled" to true.
  I think it should be called at the beginning of a transaction to process the catchup event,  then call the function "DisableCatchupInterrupt" to set "catchupInterruptEnabled" to be false to disable the signal handler to process the catchup event while the main process is doing something to avoid current issues.

xuebinsu: Then maybe we should call `AcceptInvalidationMessages()` manually.

PR #218, opened 2022-04-29
Report error if diskquota is not ready

xuebinsu:
Previously, setting quotas when diskquota is not ready is allowed.
But in that case, quotas set will not take effect. This can confuse
the user.

This patch fixes this issue by reporting an error when setting quotas
if diskquota is not ready. This is done by checking diskquota.state.

This patch also fixes possible deadlocks in the existing test cases.


quotamodel.c
```diff
@@ -565,76 +588,46 @@ check_diskquota_state_is_ready(void)
...
-		bool      isnull;
+	HeapTuple tup = SPI_tuptable->vals[0];
+	Datum     dat;
+	int       state;
+	bool      isnull;
 
-		dat = SPI_getbinval(tup, tupdesc, 1, &isnull);
-		if (isnull) continue;
-		state = DatumGetInt64(dat);
+	dat   = SPI_getbinval(tup, tupdesc, 1, &isnull);
```
beeender, line 623: The `if (isnull)` is removed, it would be good to have `NOT NULL` to the `state` column. But that requires to change the update SQL, which is a boring work...

And unlikely that could happen unless user change that. To avoid null pointer issues, I think below would be acceptable:

```
state = isnull ? DISKQUOTA_UNKNOWN_STATE : DatumGetInt32(dat);
```

beeender (commented):
LGTM Except one comment

quotamodel.c
xuebinsu: Makes sense.

beeender (approved):

Sasasu (approved):

liuxueyang (approved):

PR #220, opened 2022-05-13
Combine two object_access_hook into one

fanfuxiaoran:


diskquota_utility.c
```diff
@@ -110,6 +108,7 @@ static bool   to_delete_quota(QuotaType type, int64 quota_limit_mb, float4 segra
 static void   check_role(Oid roleoid, char *rolname, int64 quota_limit_mb);
 
 List *get_rel_oid_list(void);
+void  diskquota_stop_worker(void);
```
beeender, line 111: this is declared in header file, no need to be declared here. Actually I think compiler will report a warning.

gp_activetable.c
```diff
@@ -181,12 +181,15 @@ object_access_hook_QuotaStmt(ObjectAccessType access, Oid classId, Oid objectId,
 	if (prev_object_access_hook) (*prev_object_access_hook)(access, classId, objectId, subId, arg);
 
 	// if is 'drop extension diskquota'
-	if (classId == ExtensionRelationId)
+	if (classId == ExtensionRelationId && access == OAT_DROP)
```
beeender, line 184: the function is doing much more than `active_table`  in this file. Shall we consider moving the hook to an independent file, and expose methods from `gp_activetable` for better code organizing?

But it is not the scope of this PR.

beeender (commented):
Nice! Code deletion!!

Please don't merge this before the 2.0.0 release.

diskquota_utility.c
fanfuxiaoran: OK!

gp_activetable.c
fanfuxiaoran: Agree. We can put all the hooks into one separated file.

beeender (approved):
Merge after `2.0.0` release

gp_activetable.c
```diff
@@ -188,10 +188,7 @@ object_access_hook_QuotaStmt(ObjectAccessType access, Oid classId, Oid objectId,
 			invalidate_database_rejectmap(MyDatabaseId);
 		}
 
-		if (IS_QUERY_DISPATCHER())
```
xuebinsu: Why removing this? I don't think we have any worker on segments

xuebinsu (commented):
LGTM overall

gp_activetable.c
fanfuxiaoran:
```
update_diskquota_db_list(MyDatabaseId, HASH_REMOVE);
```
This is called in diskquota_stop_worker, which will clean monitored db cache on all segments.

xuebinsu: Then should we move `update_diskquota_db_list()` outside of `diskquota_stop_worker()`?

fanfuxiaoran: I don't think so. It's good to call update_diskquota_db_list() in diskquota_stop_worker().

Sasasu (approved):
`diskquota_stop_worker` also remote the dbid cache. fine but a little multi-usage function.

gp_activetable.c
beeender: `IS_QUERY_DISPATCHER()` was added back? cannot see the change anymore :confused:

PR #221, opened 2022-05-27
Release pipeline is publishing wrong intermediates

beeender:
"get" without "passed" will retrieve the latest content from a resource
that is not expected. Especially the release and dev pipelines share
the intermediates bucket.

- Move release intermediates to its own bucket
- Add passed the "get" step in the "exit_release" job, to make sure the
  resource version generated from previous job will be consumed in this
  job.

### DIFF
--- a/concourse/pipeline/job_def.lib.yml
+++ b/concourse/pipeline/job_def.lib.yml
@@ -1,6 +1,14 @@
 #@ load("base.lib.yml", "add_res_by_conf", "add_res_by_name")
 #@ load("@ytt:template", "template")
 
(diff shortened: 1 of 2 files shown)


xuebinsu (approved):

Sasasu (approved):
ok

debug pipeline with release build has an effect to release process.
and release build is a hidden option in `fly.sh`

🤔, I don't have a better way.

beeender:
@Sasasu 

That is actually my mistake, the `get` step has a param `passed` which will lock the resource version in current job. I missed that before.

https://concourse-ci.org/get-step.html

PR #225, opened 2022-07-25
Fix compiler version on concourse

beeender:
cmake will look up c compiler by checking /usr/bin/cc. On CentOS7, that
symbolic linked to the gcc in the same directory. Although a newer gcc
is installed, cmake will not use the newer one.

We always export the CC and CXX, then the gcc/g++ in the $PATH will be
used.

### DIFF
--- a/concourse/scripts/build_diskquota.sh
+++ b/concourse/scripts/build_diskquota.sh
@@ -6,9 +6,9 @@ function pkg() {
     [ -f /opt/gcc_env.sh ] && source /opt/gcc_env.sh
     source /usr/local/greenplum-db-devel/greenplum_path.sh
 
-    if [ "${DISKQUOTA_OS}" = "rhel6" ]; then
-        export CC="$(which gcc)"
-    fi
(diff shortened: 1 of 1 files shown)


beeender:
> Shall we add this to the cmake files, e.g. gpdb.cmake, so that other components won't suffer from this issue again?

The problematic part is the package of `gcc-4.8.5` supplies a symobilc link `/usr/bin/cc`

```
[root@3a57872e-9c27-466f-7b94-f8cf84291009 32fb0a1a]# rpm -qf /usr/bin/cc
gcc-4.8.5-44.el7.x86_64
```

But the `/opt/gcc-6.4.0/bin` doesn't have one.

So due to the priority of cmake, it finds the `cc` first, then just uses it.

Putting into `gpcb.cmake` may not be proper since the compiler is not a hard dependency of our project.

Sasasu (approved):
gpdb for centos7 is built by gcc 6.4.0, fixed by this PR.

I only notice there is a build error on centos6, so I add `if [ "${DISKQUOTA_OS}" = "rhel6" ];`
didn't notice centos7 has the same problem.

PR #226, opened 2022-08-01
[CI] Fetch secrets from vault.

higuoxing:
This patch teaches concourse to fetch secrets from vault.

### DIFF
--- a/concourse/fly.sh
+++ b/concourse/fly.sh
@@ -108,7 +108,6 @@ set -v
     sp \
     -p "${pipeline_name}" \
     -c "${yml_path}" \
-    -v "gcs-bucket=pivotal-gpdb-concourse-resources-prod" \
     -v "diskquota-branch=${branch}"
 set +v
 
--- a/concourse/fly.sh
+++ b/concourse/fly.sh
@@ -108,9 +108,7 @@ set -v
     sp \
     -p "${pipeline_name}" \
     -c "${yml_path}" \
-    -l "${workspace}/gp-continuous-integration/secrets/gpdb_common-ci-secrets.yml" \
-    -l "${workspace}/gp-continuous-integration/secrets/gp-extensions-common.yml" \
-    -l "${workspace}/gp-continuous-integration/secrets/gpdb_6X_STABLE-ci-secrets.prod.yml" \
+    -l "${workspace}/gpdb6/concourse/vars/common_prod.yml" \
     -v "diskquota-branch=${branch}"
 set +v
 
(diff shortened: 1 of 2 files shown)


concourse/pipeline/res_def.yml
```diff
@@ -21,23 +21,23 @@ resources:
 - name: diskquota_pr
   type: pull-request
   # We should rely on the webhook. See README if webhook doesn't work
-  webhook_token: ((diskquota-webhook-token))
+  webhook_token: ((extension/diskquota-webhook-token))
```
beeender, line 24: How does concourse know where is the vault server?

concourse/fly.sh
```diff
@@ -108,9 +108,7 @@ set -v
     sp \
     -p "${pipeline_name}" \
     -c "${yml_path}" \
-    -l "${workspace}/gp-continuous-integration/secrets/gpdb_common-ci-secrets.yml" \
-    -l "${workspace}/gp-continuous-integration/secrets/gp-extensions-common.yml" \
-    -l "${workspace}/gp-continuous-integration/secrets/gpdb_6X_STABLE-ci-secrets.prod.yml" \
+    -l "${workspace}/gpdb6/concourse/vars/common_prod.yml" \
```
beeender, line 111: hmmm, what is needed from this file? `gpdb6` doesn't feel common for most of us

concourse/fly.sh
```diff
@@ -108,9 +108,7 @@ set -v
     sp \
     -p "${pipeline_name}" \
     -c "${yml_path}" \
-    -l "${workspace}/gp-continuous-integration/secrets/gpdb_common-ci-secrets.yml" \
```
fanfuxiaoran, line 111: The values of the vars in the pipeline are saved in gpdb repo?

higuoxing: No, it's a private repo.

fanfuxiaoran:
```
 -l "${workspace}/gpdb6/concourse/vars/common_prod.yml"
```
It looks like in the gpdb6 repo, the name should be changed, a little confusing.

concourse/fly.sh
higuoxing: `gcs-bucket` is needed in our pipeline.

concourse/fly.sh
higuoxing: Yes, there are some commonly used variables stored in that file and it's public. There's no secret in that file.

beeender: > -l https://github.com/greenplum-db/gpdb/blob/master/concourse/vars/common_prod.yml

This actually looks nice! Does this work?

beeender: `-l` doesn't support http :(

beeender (approved):
:clap:

higuoxing:
CI is updated.

PR #227, opened 2022-08-15
Ignore cmake files for git.

zhrt123:


higuoxing (approved):

beeender:
do:

```
mkdir build
cd build
cmake ..
make
```

Then all the intermediate files will only be in `build` directory, and they have been ignored already. Plus that is the convention of cmake.

Sasasu:
BTW, you can add those to `~/.gitignore`

this is my `~/.gitignore`

```
.DS_Store
.ccls-cache
compile_commands.json
.idea/*
build/*
data/*
third-party/*
build/*
.vscode/*
```

PR #228, opened 2022-08-16
Fix bug: diskquota quota usage doesn't include uncommitted table

zhrt123:
Modify the show_fast_schema_quota_view, show_fast_role_quota_view, show_fast_schema_tablespace_quota_view and show_fast_role_tablespace_quota_view. Combine pg_class with relation_cache into all_relation, and join all_relation with diskquota.table_size to calculate quota usage.


higuoxing:
Seems that CI is failed with:

```
=======================
 1 of 12 tests failed. 
=======================

The differences that caused some tests to fail can be viewed in the
file "/tmp/build/61170088/bin_diskquota/tests/isolation2/regression.diffs".  A copy of the test summary that you see
above is saved in the file "/tmp/build/61170088/bin_diskquota/tests/isolation2/regression.out".

======================================================================
DIFF FILE: /home/gpadmin/diskquota_artifacts/tests/isolation2/regression.diffs
======================================================================

--- \/tmp\/build\/61170088\/bin_diskquota\/tests\/isolation2\/expected\/test_fast_quota_view\.out	2022-08-17 07:27:51.456468751 +0000
+++ \/tmp\/build\/61170088\/bin_diskquota\/tests\/isolation2\/results\/test_fast_quota_view\.out	2022-08-17 07:27:51.460469112 +0000
@@ -119,7 +119,7 @@
 2: SELECT role_name, quota_in_mb, rolsize_in_bytes FROM diskquota.show_fast_role_quota_view;
  role_name | quota_in_mb | rolsize_in_bytes 
 -----------+-------------+------------------
- r         | 100         | 7864320          
+ r         | 100         | 6488064          
 (1 row)
 2: SELECT role_name, tablespace_name, quota_in_mb, rolsize_tablespace_in_bytes FROM diskquota.show_fast_role_tablespace_quota_view;
  role_name | tablespace_name | quota_in_mb | rolsize_tablespace_in_bytes 

======================================================================
gmake[3]: *** [tests/CMakeFiles/isolation2] Error 1
gmake[2]: *** [tests/CMakeFiles/isolation2.dir/all] Error 2
gmake[1]: *** [tests/CMakeFiles/installcheck.dir/rule] Error 2
gmake: *** [installcheck] Error 2
```

diskquota--2.0--2.1.sql
```diff
@@ -0,0 +1,239 @@
...
+  relation_cache AS (
+    SELECT (f).* FROM diskquota.show_relation_cache() as f
+  ),
+  all_relation AS (
+    SELECT DISTINCT(oid), relowner, relnamespace, reltablespace from (
+      SELECT relid as oid, owneroid as relowner, namespaceoid as relnamespace, spcnode as reltablespace FROM relation_cache
+      UNION
+      SELECT oid, relowner, relnamespace, reltablespace from pg_class
+    ) as union_relation
+  ),
```
fanfuxiaoran, line 89: It seems that packing those lines into one view or function is better. Then other views like `show_fast_role_quota_view` can use directly.

zhrt123: Good idea!

diskquota--2.1.sql
```diff
@@ -0,0 +1,318 @@
...
+		SELECT diskquota.show_relation_cache() AS a
+		FROM  gp_dist_random('gp_id')
+	)
+	SELECT (a).* FROM relation_cache; $$ LANGUAGE SQL;
+
+-- view part
+CREATE VIEW diskquota.show_all_relation_view AS
+WITH
+  relation_cache AS (
+    SELECT (f).* FROM diskquota.show_relation_cache() as f
```
fanfuxiaoran, line 120: Will this view be run on all segs or just on the master? Should use `show_relation_cache_all_seg` instead of `show_relation_cache` here?

zhrt123: I think it will just be run in master. Are any scenes where the `show_fast_schema_quota_view` is run in segments previously?

fanfuxiaoran: I just wonder  show_relation_cache` only runs on master,if we can collect all the relation oid caches as some are recorded on segments.

zhrt123: > I just wonder show_relation_cache` only runs on master,
It should only run on master.

fanfuxiaoran: > It should only run on master.

Can it collect all the relation oid caches as some are recorded on segments?

zhrt123: The oid which is recorded in segments will also be recorded in master, so we just need to collect the relation oid cache on master.

fanfuxiaoran: Like `active_table_hook_smgrextend`, does this hook also run both on segments and master?

zhrt123: It is used to test `relation_cache` in regression, so we can not remove this function.

fanfuxiaoran: I think it's useless, and no need to test it, just testing `show_relation_cache` enough.

zhrt123: `show_relation_cache` can not reach the `relation_cache` on segments.

fanfuxiaoran (approved):

PR #233, opened 2022-08-29
Remove free writer gang in launcher

fanfuxiaoran:
As launcher and worker both run in bgworker, they are clients when
connecting to segments. They know when should to close the connections.
So we directly call `DisconnectAndDestroyAllGangs` to close them and
remove the code that only closes free read gang based on timeout.

For launcher, we destroy gangs after init_datbase_list or after each
time processing create/drop diskquota extension.

For worker, we only destroy gangs after the extension is paused.


higuoxing (approved):
Looks good!

diskquota.c
```diff
@@ -530,6 +531,12 @@ disk_quota_worker_main(Datum main_arg)
 			/* Refresh quota model with init mode */
 			refresh_disk_quota_model(!MyWorkerInfo->dbEntry->inited);
 			MyWorkerInfo->dbEntry->inited = true;
+			is_gang_destroyed             = false;
+		}
+		else if (!is_gang_destroyed)
```
zhrt123, line 536:
```suggestion
		if (!is_gang_destroyed)
```
Can we change `else if` to `if`? Because it should be triggered in the next loop if `else if` is used.

fanfuxiaoran: > Can we change else if to if? Because it should be triggered in the next loop if else if is used.
No, if the diskquota is not paused, we would like to reuse the gang.

zhrt123: Why we should destroy all gangs when `diskquota is not paused` and `gang is not destroyed`?

fanfuxiaoran: > Why we should destroy all gangs when diskquota is not paused and gang is not destroyed?

We destroy gangs when `diskquota is paused and gang is not destoried`

The user can pause diskquota and resume it, when diskquota is paused, the diskquota worker process still runs, but does nothing, so we just destroy the gang.  And if the gang is closed, there is no need to close it again.

zhrt123: OK, I have mistaken it, now it is clear.

zhrt123 (approved):
LGTM

PR #236, opened 2022-09-15
Fix the test failure on release pipeline

fanfuxiaoran:
Fix test failure on release pipeline
Regress tests failed on the release
pipeline due to the diskquota.naptime
is not 0. As diskquota launcher
first sleeps, then starts diskquota
workers. When naptime is not 0, workers
are started very late, then some tests
which checking them failed.

In this pr, fix the test by starting
diskquota workers in the launcher before
it sleeps. This makes diskquota workers
will be started immediately when the launcher
starts.

Also, increased the sleep time in
the tests.

Using get_database_name instead
of get_db_name, as worker_get_epoch
is called in a transaction.

### DIFF
--- a/diskquota.c
+++ b/diskquota.c
@@ -1364,7 +1364,7 @@ worker_get_epoch(Oid dbid)
 	if (!found)
 	{
 		ereport(ERROR, (errcode(ERRCODE_INTERNAL_ERROR),
-		                errmsg("[diskquota] worker not found for database \"%s\"", get_db_name(dbid))));
+		                errmsg("[diskquota] worker not found for database \"%s\"", get_database_name(dbid))));
 	}
 	return epoch;
 }
--- a/diskquota.c
+++ b/diskquota.c
@@ -1363,7 +1363,7 @@ worker_get_epoch(Oid dbid)
 	LWLockRelease(diskquota_locks.monitored_dbid_cache_lock);
 	if (!found)
 	{
-		ereport(ERROR, (errcode(ERRCODE_INTERNAL_ERROR),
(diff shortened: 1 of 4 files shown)


Sasasu (approved):

Sasasu (approved):

diskquota.c
```diff
@@ -640,15 +640,52 @@ disk_quota_launcher_main(Datum main_arg)
...
+			if (nap.tv_sec == 0 && nap.tv_usec == 0)
+			{
+				loop_start_time = GetCurrentTimestamp();
+				/* set the curDB pointing to the head of the db list */
+				curDB = NULL;
+			}
+			/* do nothing, just to sleep untill the nap time is 0 */
+			else
+			{
+				continue;
```
higuoxing, line 670:
```suggestion
				pg_sleep(0.1);
				continue;
```

What about adding `pg_sleep(little_amount_of_time)` to this loop to reduce some CPU consumption?

higuoxing (commented):
Looks good!

diskquota.c
fanfuxiaoran: > What about adding `pg_sleep(little_amount_of_time)` to this loop to reduce some CPU consumption?

Hmm... this looks not helpful to resolve the issue.

higuoxing: I mean if the naptime is 1s and `curDB == DiskquotaLauncherShmem->dbArrayTail` (the launcher have finished one loop), the launcher will keep looping here and do nothing, the load of the CPU that occupied by the launcher process will be 100% for 1s.

I don't insist on this and please correct me if I'm wrong.

higuoxing (approved):

PR #239, opened 2022-10-13
Support at least 20000 tables for each database

zhrt123:
Currently, diskquota supports max to 8000 tables, including partition tables. It may cause unexpected results if there are too many tables in the customer cluster.

This PR change `init_size` of table_size_map to 20000, instead of MAX_TABLES, so that we can
economize the actual usage of SHM. To handle scenarios with a large number of tables,
we set `max_size` to MAX_TABLES (1000000).

To avoid memory waste, only the master should be allocated shared memory for diskquota_launcher and diskquota_worker.

### DIFF
--- a/quotamodel.c
+++ b/quotamodel.c
@@ -475,8 +475,13 @@ DiskQuotaShmemSize(void)
 	size = add_size(size, hash_estimate_size(diskquota_max_active_tables, sizeof(Oid)));
 	size = add_size(size, hash_estimate_size(MAX_NUM_MONITORED_DB,
 	                                         sizeof(struct MonitorDBEntryStruct))); // monitored_dbid_cache
-	size = add_size(size, diskquota_launcher_shmem_size());
-	size = add_size(size, diskquota_worker_shmem_size() * MAX_NUM_MONITORED_DB);
+
+	if (IS_QUERY_DISPATCHER())
+	{
+		size = add_size(size, diskquota_launcher_shmem_size());
+		size = add_size(size, diskquota_worker_shmem_size() * MAX_NUM_MONITORED_DB);
+	}
+
 	return size;
 }
 
--- a/quotamodel.c
+++ b/quotamodel.c
@@ -44,7 +44,7 @@
 
 /* cluster level max size of rejectmap */
 #define MAX_DISK_QUOTA_REJECT_ENTRIES (1024 * 1024)
-#define MAX_TABLES (1024L * 8)
+#define MAX_TABLES (20 * 1024)
 /* cluster level init size of rejectmap */
 #define INIT_DISK_QUOTA_REJECT_ENTRIES 8192
 /* per database level max size of rejectmap */
@@ -68,17 +68,23 @@ int SEGCOUNT = 0;
  */
 struct TableSizeEntry
 {
-	Oid   reloid;
-	int16 segid;
-	Oid   tablespaceoid;
-	Oid   namespaceoid;
-	Oid   owneroid;
+	Oid    reloid;
+	int    segid;
+	Oid    tablespaceoid;
+	Oid    namespaceoid;
+	Oid    owneroid;
+	uint32 flag;     /* flag's each bit is used to show the table's status,
+	                  * which is described in TableSizeEntryFlag.
+	                  */
 	int64 totalsize; /* table size including fsm, visibility map
 	                  * etc. */
-	bool is_exist;   /* flag used to check whether table is already
-	                  * dropped */
-	bool need_flush; /* whether need to flush to table table_size */
 };
+
+typedef enum
+{
+	TABLE_EXIST      = (1 << 0), /* whether table is already dropped */
+	TABLE_NEED_FLUSH = (1 << 1)  /* whether need to flush to table table_size */
+} TableSizeEntryFlag;
 
 /*
  * table disk size and corresponding schema and owner
@@ -193,6 +199,10 @@ static void init_lwlocks(void);
 static void export_exceeded_error(GlobalRejectMapEntry *entry, bool skip_name);
 void        truncateStringInfo(StringInfo str, int nchars);
 static void format_name(const char *prefix, uint32 id, StringInfo str);
+
+static bool get_table_size_entry_flag(TableSizeEntry *entry, TableSizeEntryFlag flag);
+static void reset_table_size_entry_flag(TableSizeEntry *entry, TableSizeEntryFlag flag);
+static void set_table_size_entry_flag(TableSizeEntry *entry, TableSizeEntryFlag flag);
 
 /* add a new entry quota or update the old entry quota */
 static void
@@ -440,7 +450,7 @@ static Size
 diskquota_worker_shmem_size()
 {
 	Size size;
-	size = hash_estimate_size(1024 * 1024, sizeof(TableSizeEntry));
+	size = hash_estimate_size(MAX_TABLES, sizeof(TableSizeEntry));
 	size = add_size(size, hash_estimate_size(MAX_LOCAL_DISK_QUOTA_REJECT_ENTRIES, sizeof(LocalRejectMapEntry)));
 	size = add_size(size, hash_estimate_size(1024L, sizeof(struct QuotaMapEntry)) * NUM_QUOTA_TYPES);
 	return size;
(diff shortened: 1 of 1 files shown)


quotamodel.c
```diff
@@ -68,18 +68,25 @@ int SEGCOUNT = 0;
...
-	int16 segid;
-	Oid   tablespaceoid;
-	Oid   namespaceoid;
-	Oid   owneroid;
+	Oid    reloid;
+	int    segid;
+	Oid    tablespaceoid;
+	Oid    namespaceoid;
+	Oid    owneroid;
+	uint32 flag;     /* each bit in flag is used to show table's status
```
higuoxing, line 76: What about using `uint8_t`, since we only need 2 bits here? Also, the comment is a liitle bit confusing, let's represent these flags in binary format? e.g.,

```c
/*
 * 0b0000000[bit]: Represent whether the table has been dropped.
 * 0b000000[bit]0: Represent whether the table needs to be flushed to the table_size table.
 */
```

quotamodel.c
```diff
@@ -68,18 +68,25 @@ int SEGCOUNT = 0;
...
-	bool is_exist;   /* flag used to check whether table is already
-	                  * dropped */
-	bool need_flush; /* whether need to flush to table table_size */
 };
 
+typedef enum
+{
+	TABLE_EXIST,
+	NEED_FLUSH
+} TableSizeEntryFlag;
```
higuoxing, line 88: It might be better to use macros to represent these masks.

```c
#define TABLE_EXITS (1 << 0) // Indicates whether the given table exists.
#define TABLE_NEEDS_FLUSH (1 << 1) // Indicates whether the given table needs flush to table_size table.
```

and then, we can rewrite these functions

```c
static bool
get_table_size_entry_flag(TableSizeEntry *entry, TableSizeEntryFlag flag)
{
	return (entry->flag & (1 << flag)) ? true : false;
}

static void
reset_table_size_entry_flag(TableSizeEntry *entry, TableSizeEntryFlag flag)
{
	if (entry->flag & (1 << flag)) entry->flag ^= (1 << flag);
}

static void
set_table_size_entry_flag(TableSizeEntry *entry, TableSizeEntryFlag flag)
{
	entry->flag |= (1 << flag);
}
```

to

```c
static bool
get_table_size_entry_flag(TableSizeEntry *entry, uint8_t mask)
{
	return entry->flag & mask;
}

static void
reset_table_size_entry_flag(TableSizeEntry *entry, uint8_t mask)
{
	entry->flag &= ~mask;
}

static void
set_table_size_entry_flag(TableSizeEntry *entry, uint8_t mask)
{
	entry->flag |= mask;
}
```

quotamodel.c
zhrt123: The struct is aligned by 4byte, so it it not necessary to use uint8 for `flag`

higuoxing: > The struct is aligned by 4byte, so it is not necessary to use uint8 for `flag`

Makes sense.

quotamodel.c
zhrt123: I think `enum` is better to present the flag's meaning. But `entry->flag &= ~mask` is a nice change, I will take it.

quotamodel.c
zhrt123: And int64 is aligned by 8byte, so I move it to the bottom of the struct.

zhrt123: > the comment is a liitle bit confusing

I have updated the note of `flag` in commit 88376d9836dbcda7ce6e8f7dd1b7d3d48362f1f8.

quotamodel.c
higuoxing: The shortage of using `enum` type is that we cannot specify multiple flags for `entry->flag`, though we don't need that currently.

With macros, you can do

```c
set_table_size_entry_flag(entry, TABLE_EXITS | TABLE_NEEDS_FLUSH);
```

while you cannot do that with `enum` type.

fanfuxiaoran: > The shortage of using `enum` type is that we cannot specify multiple flags for `entry->flag`, though we don't need that currently.
> 
> With macros, you can do
> 
> ```c
> set_table_size_entry_flag(entry, TABLE_EXITS | TABLE_NEEDS_FLUSH);
> ```
> 
> while you cannot do that with `enum` type.

I prefer macros too, as using `enum` to define flags is not such traditional.

zhrt123: Indeed many state lists, type lists, and flag lists are defined as an enum type in diskquota, I prefer to define this flag list as an enum type.

quotamodel.c
```diff
@@ -68,18 +71,24 @@ int SEGCOUNT = 0;
...
 	                  * etc. */
-	bool is_exist;   /* flag used to check whether table is already
-	                  * dropped */
-	bool need_flush; /* whether need to flush to table table_size */
 };
 
+typedef enum
+{
+	TABLE_EXIST      = (1 << 0), /* whether table is already dropped */
+	TABLE_NEED_FLUSH = (1 << 1)  /* whether need to flush to table table_size */
```
Sasasu, line 89: C98 support `0b01` and `0b10`

Sasasu (approved):

quotamodel.c
Sasasu:
```C
#include <stdint.h>

struct the_old {
	int32_t reloid;
	int16_t segid;
	int32_t tablespaceoid;
	int32_t namespaceoid;
	int32_t owneroid;
	int64_t totalsize;
	bool    is_exist;
	bool    need_flush;
};

sizeof = 40

---

struct the_new {
	int32_t  reloid;
	int      segid;
	int32_t  tablespaceoid;
	int32_t  namespaceoid;
	int32_t  owneroid;
	uint32_t flag;
	int64_t  totalsize;
};

sizeof = 32 (20% better)

---

struct the_magic {
	int32_t  reloid;
	int      segid;
	int32_t  tablespaceoid;
	int32_t  namespaceoid;
	int32_t  owneroid;
	int64_t  totalsize;
	uint8_t flag;
} __attribute__((packed)) ;

sizeof = 29 (27.5% better)
```

we can remove the memory align using __attribute___((packed)). I think there will be no performance degradation in this case.

quotamodel.c
```diff
@@ -2103,3 +2122,21 @@ format_name(const char *prefix, uint32 id, StringInfo str)
 	appendStringInfo(str, "%s_%u", prefix, id);
 	Assert(str->len <= SHMEM_INDEX_KEYSIZE);
 }
+
+static bool
+get_table_size_entry_flag(TableSizeEntry *entry, TableSizeEntryFlag flag)
+{
+	return (entry->flag & flag) ? true : false;
```
fanfuxiaoran, line 2129:
```suggestion
	return entry->flag & flag;
```

quotamodel.c
fanfuxiaoran: > we can remove the memory align using attribute_((packed)). I think there will be no performance degradation in this case.

Can you explain more why not packing here will not lead to performance issues?
 And unpacking here has no effect. 
`elementSize = MAXALIGN(sizeof(HASHELEMENT)) + MAXALIGN(hctlv->entrysize);`  hash function will align it anyway.

Sasasu: > Can you explain more why not packing here will not lead to performance issues?

the unaligned memory access will never be a performance issue in any PostgreSQL program.
the other part is slower than the memory access.

and our use case for calculation:
```
<register> =  <memory>
<r> = SPI_get() <-- a very very long journey. break all register and L1 cache 
<register> = add <register> <r>
<memory> = <register>
```

consider micro arch performance is useless in PostgreSQL.

and in this case, a memory bottleneck case. save 27% L3 cache is faster than double the L1 access.


> And unpacking here has no effect.

Ok, this is due to the lousy hash map library.

quotamodel.c
zhrt123: But we need to add `0` in previous flags when adding a new flag.

Sasasu: ok, never maind.

quotamodel.c
fanfuxiaoran: I think there is an Implicit type cast here to cast uint32 to bool

zhrt123: In GPDB, there is no type named `bool` and it is defined by `char`. Unlike `if`, the function returning value cannot be cast to `true` or `false`.

fanfuxiaoran: Hmm... bool is defined as char.  Understand why you said there is an overflow here. Thanks.

fanfuxiaoran (approved):

PR #240, opened 2022-10-18
Fix flaky test

fanfuxiaoran:


Sasasu (approved):

beeender (approved):

tests/regress/expected/test_extension.out
```diff
@@ -59,7 +59,7 @@ show diskquota.max_workers;
  20
 (1 row)
 
-\! sleep 1.5; ps -ef | grep postgres | grep "\[diskquota]" | grep -v grep | wc -l
+\! sleep 3; ps -ef | grep postgres | grep "\[diskquota]" | grep -v grep | wc -l
```
higuoxing, line 62: Maybe we can replace it with a while loop? I'm not good at writing bash scripts

```
while [ $(ps -ef | grep postgres | grep "\[diskquota]" | grep -v grep | wc -l) -ne 3 ]; do sleep 0.5; done
```

fanfuxiaoran: A good idea, a while loop with limited times, I will try it later.

higuoxing: :rofl: Never mind, my comments come too late.

PR #241, opened 2022-10-18
Fix bug: relation_cache cannot isolate table info for each database

zhrt123:
Previously, users will see the information of tables in db2 when accessing db1 and executing `select diskquota.show_relation_cache()`.  Meanwhile, if db1's bgworker can see the table in db2, the performance will be influenced, because the table size in db2 will be calculated although it is not useful for db1 diskquota.

This PR has added a judgment to prevent this situation.

### DIFF
--- a/gp_activetable.c
+++ b/gp_activetable.c
@@ -791,6 +791,9 @@ get_active_tables_oid(void)
 		RelFileNode rnode;
 		Oid         prelid;
 
+		/* The session of db1 should not see the table inside db2. */
+		if (active_table_file_entry->dbid != MyDatabaseId) continue;
+
 		rnode.dbNode  = active_table_file_entry->dbid;
 		rnode.relNode = active_table_file_entry->relfilenode;
 		rnode.spcNode = active_table_file_entry->tablespaceoid;
--- a/quotamodel.c
+++ b/quotamodel.c
@@ -793,7 +793,8 @@ merge_uncommitted_table_to_oidlist(List *oidlist)
 	hash_seq_init(&iter, relation_cache);
 	while ((entry = hash_seq_search(&iter)) != NULL)
 	{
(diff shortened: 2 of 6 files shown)


higuoxing:
Can we test this behavior? We can connect to different databases via the following snippet

```
\c dbname
```

zhrt123:
> Can we test this behavior? We can connect to different databases via the following snippet
> 
> ```
> \c dbname
> ```

Yes, we can connect to different db via this instruction. However, in diskquota, if db1's bgworker can see the table in db2, the performance will be influenced. db2's table size is not useful for db1 diskquota, but these table sizes should always be calculated.

higuoxing:
> > Can we test this behavior? We can connect to different databases via the following snippet
> > ```
> > \c dbname
> > ```
> 
> Yes, we can connect to different db via this instruction. However, in diskquota, if db1's bgworker can see the table in db2, the performance will be influenced. db2's table size is not useful for db1 diskquota, but these table sizes should always be calculated.

Sorry for the misleading, I mean can we add a test case for this change? E.g., connecting to different database and check the output of `diskquota.show_relation_cache()`.

zhrt123:
> > > Can we test this behavior? We can connect to different databases via the following snippet
> > > ```
> > > \c dbname
> > > ```
> > 
> > 
> > Yes, we can connect to different db via this instruction. However, in diskquota, if db1's bgworker can see the table in db2, the performance will be influenced. db2's table size is not useful for db1 diskquota, but these table sizes should always be calculated.
> 
> Sorry for the misleading, I mean can we add a test case for this change? E.g., connecting to different database and check the output of `diskquota.show_relation_cache()`.

Good idea.

fanfuxiaoran (approved):

higuoxing (approved):
Looks good!

PR #243, opened 2022-10-20
Reduce MAX_TABLES to 200K.

zhrt123:
The scale of 50 databases with 200K tables is large enough, so we reduce MAX_TABLES to 200K to avoid memory waste.


beeender (approved):
LGTM

Need @radarwave confirm

radarwave (approved):
LGTM

PR #245, opened 2022-10-26
Rewrite the diskquota worker scheduler

fanfuxiaoran:
Rewrite the diskquota worker scheduler. Each db entry manages the running time by itself to make the scheduler more simple.
I have tested the performance, nearly the same as before.
Add a  UDF db_status to help debug or monitor the diskquota, related SQL is in test_util.sql. I didn't put it into diskquota--2.1.sql, because I don't want to users to use it now. I didn't use any shared lock when getting those status info, so they may be not consistent .

### DIFF
--- a/diskquota.c
+++ b/diskquota.c
@@ -1787,9 +1787,10 @@ next_db(DiskquotaDBEntry *curDB)
 		if (nextSlot >= MAX_NUM_MONITORED_DB) nextSlot = 0;
 		DiskquotaDBEntry *dbEntry = &DiskquotaLauncherShmem->dbArray[nextSlot];
 		nextSlot++;
-		if (!dbEntry->in_use) continue;
-		if (dbEntry->workerId != INVALID_WORKER_ID) continue;
-		if (dbEntry->dbid == InvalidOid) continue;
+		if (!dbEntry->in_use ||
+		    dbEntry->workerId != INVALID_WORKER_ID ||
+		    dbEntry->dbid == InvalidOid)
+			continue;
 		result = dbEntry;
 		break;
 	}
--- a/diskquota.c
+++ b/diskquota.c
@@ -86,7 +86,7 @@ static int num_db = 0;
 // in shared memory, only for launcher process
 DiskquotaLauncherShmemStruct *DiskquotaLauncherShmem;
 
-#define MIN_SLEEPTIME 100.0 /* milliseconds */
+#define MIN_SLEEPTIME 100 /* milliseconds */
 
 /*
  * bgworker handles, in launcher local memory,
@@ -1349,7 +1349,7 @@ start_worker(DiskquotaDBEntry *dbEntry)
 	}
 
 	Assert(status == BGWH_STARTED);
-	dbEntry->status = RUNNING;
+	dbEntry->status = SLOT_RUNNING;
 	return true;
 Failed:
 
@@ -1622,7 +1622,7 @@ FreeWorker(DiskQuotaWorkerEntry *worker)
 			if (in_use && worker->dbEntry->workerId == worker->id)
 			{
 				worker->dbEntry->workerId = INVALID_WORKER_ID;
-				worker->dbEntry->status   = SLEEPING;
+				worker->dbEntry->status   = SLOT_SLEEPING;
 				worker->dbEntry->next_run_time =
 				        TimestampTzPlusMilliseconds(GetCurrentTimestamp(), diskquota_naptime * 1000L);
 			}
@@ -1688,7 +1688,7 @@ init_launcher_shmem()
 			memset(&DiskquotaLauncherShmem->dbArray[i], 0, sizeof(DiskquotaDBEntry));
 			DiskquotaLauncherShmem->dbArray[i].id       = i;
 			DiskquotaLauncherShmem->dbArray[i].workerId = INVALID_WORKER_ID;
-			DiskquotaLauncherShmem->dbArray[i].status   = UNUSED;
+			DiskquotaLauncherShmem->dbArray[i].status   = SLOT_UNUSED;
 		}
 	}
 }
@@ -1716,7 +1716,7 @@ add_db_entry(Oid dbid)
 			dbEntry->dbid          = dbid;
 			dbEntry->in_use        = true;
 			dbEntry->next_run_time = GetCurrentTimestamp();
-			dbEntry->status        = SLEEPING;
+			dbEntry->status        = SLOT_SLEEPING;
 			result                 = dbEntry;
 		}
 		else if (dbEntry->in_use && dbEntry->dbid == dbid)
@@ -1787,10 +1787,7 @@ next_db(DiskquotaDBEntry *curDB)
 		if (nextSlot >= MAX_NUM_MONITORED_DB) nextSlot = 0;
 		DiskquotaDBEntry *dbEntry = &DiskquotaLauncherShmem->dbArray[nextSlot];
 		nextSlot++;
-		if (!dbEntry->in_use ||
-		    dbEntry->workerId != INVALID_WORKER_ID ||
-		    dbEntry->dbid == InvalidOid)
-			continue;
+		if (!dbEntry->in_use || dbEntry->workerId != INVALID_WORKER_ID || dbEntry->dbid == InvalidOid) continue;
 		result = dbEntry;
 		break;
 	}
@@ -1854,7 +1851,7 @@ vacuum_db_entry(DiskquotaDBEntry *db)
 	db->dbid     = InvalidOid;
 	db->inited   = false;
 	db->workerId = INVALID_WORKER_ID;
-	db->status   = UNUSED;
+	db->status   = SLOT_UNUSED;
 	db->in_use   = false;
 }
 
--- a/diskquota.c
+++ b/diskquota.c
@@ -663,21 +663,17 @@ disk_quota_launcher_main(Datum main_arg)
 		}
 		else
 		{
-			long        secs;
-			int         microsecs;
 			TimestampTz curTime = GetCurrentTimestamp();
-			TimestampDifference(curTime, curDB->next_run_time, &secs, &microsecs);
-			nap.tv_sec  = secs;
-			nap.tv_usec = microsecs;
+			TimestampDifference(curTime, curDB->next_run_time, &nap.tv_sec, &nap.tv_usec);
 
 			/* if the sleep time is too short, just skip the sleeping */
-			if (secs == 0 && microsecs < MIN_SLEEPTIME * 1000L)
+			if (nap.tv_sec == 0 && nap.tv_usec < MIN_SLEEPTIME * 1000L)
 			{
 				nap.tv_usec = 0;
 			}
 
 			/* if the sleep time is too long, advance the next_run_time */
-			if (secs > diskquota_naptime)
+			if (nap.tv_sec > diskquota_naptime)
 			{
 				nap.tv_sec           = diskquota_naptime;
 				nap.tv_usec          = 0;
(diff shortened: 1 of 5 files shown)


diskquota.c
```diff
@@ -735,10 +736,48 @@ disk_quota_launcher_main(Datum main_arg)
...
+		{
+			advance_one_db = true;
+			continue;
+		}
+
+		/*
+		 * Try to start a worker to run the db if has exceeded the next_run_time
+		 * if start_worker fails, advance_one_db will be set to false, so in the
+		 * next loop will run the db again.
+		 */
```
higuoxing, line 768: Will it be possible that we start the worker for the current database, but the worker takes too much time at some point and we have to start the worker for the same database in the next iteration, then it ends up starting workers for the same database again and again ...?

diskquota.c
```diff
@@ -84,25 +84,9 @@ static DiskQuotaWorkerEntry *volatile MyWorkerInfo = NULL;
...
- * 2) when curDB is DiskquotaLauncherShmem->dbArrayTail,
- * means it had finish one loop just now. And should
- * sleep for ${diskquota.naptime} sconds.
- *
- * 3) when curDB is pointing to any db entry in
- * DiskquotaLauncherShmem->dbArray[], it means it is in
- * one loop to start each worker for each database.
- */
-static DiskquotaDBEntry *curDB = NULL;
+#define MIN_SLEEPTIME 100.0 /* milliseconds */
```
higuoxing, line 89: Why do we need a floating point number here? I think integer is good enough.

diskquota.c
fanfuxiaoran: That's a good point, I have met with that problem.  Fixed it by checking if the db is not running in the `next_db` routine.
When the `workerId` is Invalid means there is no worker is running for it, so we can avoid the issue you said.

diskquota.c
fanfuxiaoran: Because it represents milliseconds, there are some codes like ` if (secs == 0 && microsecs < MIN_SLEEPTIME * 1000L)` to change it into microsecs, so it's better to keep one decimal in it.

fanfuxiaoran: But seems we didn't make use of this feature, I will consider changing it into an integer.

diskquota_utility.c
```diff
@@ -1655,3 +1655,76 @@ check_role(Oid roleoid, char *rolname, int64 quota_limit_mb)
 		ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
 		                errmsg("Can not set disk quota for system owner: %s", rolname)));
 }
+
+PG_FUNCTION_INFO_V1(status);
+Datum
+status(PG_FUNCTION_ARGS)
```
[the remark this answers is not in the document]
fanfuxiaoran: agree with db_status()

test_util.sql
```diff
@@ -0,0 +1,21 @@
...
+	"id"   smallint,
+	"dbid" oid,
+	"workerid" smallint,
+	"status" smallint,
+	"last_run_time" timestamptz,
+	"cost" smallint,
+	"next_run_time" timestamptz,
+	"epoch" int8 
+);
+CREATE FUNCTION diskquota.db_status() RETURNS setof diskquota.db_status AS '$libdir/diskquota-2.1.so',  'status' LANGUAGE C VOLATILE;
```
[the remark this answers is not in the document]
fanfuxiaoran: I don't know any way to remove the specific diskquota version when defining UDFs. I think we need a story or issue to resolve it.

test_util.sql
```diff
@@ -0,0 +1,21 @@
...
+CREATE TYPE diskquota.db_status AS (
+	"id"   smallint,
+	"dbid" oid,
+	"workerid" smallint,
+	"status" smallint,
+	"last_run_time" timestamptz,
+	"cost" smallint,
+	"next_run_time" timestamptz,
+	"epoch" int8 
+);
```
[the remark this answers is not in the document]
fanfuxiaoran: Yep, fix it.

test_util.sql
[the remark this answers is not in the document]
zhrt123: By replacing all.

diskquota_utility.c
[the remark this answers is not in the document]
zhrt123: Making it the same as diskquota.db_status should be appropriate, such as `db_status`.

diskquota.c
```diff
@@ -699,6 +697,7 @@ disk_quota_launcher_main(Datum main_arg)
 
 		if (nap.tv_sec != 0 || nap.tv_usec != 0)
 		{
+			elog(DEBUG1, "[diskquota] naptime sec:%ld, usec:%d", nap.tv_sec, nap.tv_usec);
 			rc = WaitLatch(&MyProc->procLatch, WL_LATCH_SET | WL_TIMEOUT | WL_POSTMASTER_DEATH,
 			               (nap.tv_sec * 1000L) + (nap.tv_usec / 1000L));
 			ResetLatch(&MyProc->procLatch);
```
zhrt123, line 703: Since nap.tv_sec will be set to at least 1, can we remove this line: `if (!diskquota_naptime) usleep(1);`?

diskquota.c
```diff
@@ -638,53 +623,66 @@ disk_quota_launcher_main(Datum main_arg)
...
+				nap.tv_usec = 0;
+			}
+
+			/* if the sleep time is too long, advance the next_run_time */
+			if (secs > diskquota_naptime)
+			{
+				nap.tv_sec           = diskquota_naptime;
+				nap.tv_usec          = 0;
+				curDB->next_run_time = TimestampTzPlusMilliseconds(curTime, diskquota_naptime * 1000L);
+			}
```
zhrt123, line 685:
```suggestion
			/* if the sleep time is too short, just skip the sleeping */
			if (secs == 0 && microsecs < MIN_SLEEPTIME * 1000L)
			{
				nap.tv_usec = 1;
				nap.tv_usec = 0;
				curDB->next_run_time = TimestampTzPlusMilliseconds(curTime, 1000L);
			}

			/* if the sleep time is too long, advance the next_run_time */
			if (nap.tv_secs >= diskquota_naptime)
			{
				nap.tv_sec           = diskquota_naptime;
				nap.tv_usec          = 0;
				curDB->next_run_time = TimestampTzPlusMilliseconds(curTime, diskquota_naptime * 1000L);
			}
```
The logic of this code is a little messy. Firstly, if the sleep time is too short, we should also set `nap.tv_usec` to 1, as the comment `wait at least one time slice, avoid 100% CPU usage` is written in the 705th line. Secondly, I think it is better to describe the compare statement as `nap.tv_sec > diskquota_naptime` in the second `if` statement.

diskquota.c
fanfuxiaoran: We only do the sleep when diskquota_nap time is 0.

diskquota.c
zhrt123: Then, we can remove `if (!diskquota_naptime) usleep(1);` in the 706th line.

diskquota.c
zhrt123: If the `workerId` is valid, what will launcher do?

fanfuxiaoran: > If the workerId is valid, what will launcher do?
Skip the current db, move on to the next db. ( in next_db function);

diskquota.c
```diff
@@ -735,10 +736,48 @@ disk_quota_launcher_main(Datum main_arg)
...
+		/*
+		 * When db list is empty, curDB is NULL.
+		 * When curDB->in_use is false means dbEtnry has been romoved
+		 * When curDB->dbid doesn't equtal curDBId, it means the slot has
+		 * been used by another db
+		 *
+		 * For the above conditions, we just skip this loop and try to fetch
+		 * next db to run.
+		 */
+		if (curDB == NULL || !curDB->in_use || curDB->dbid != curDBId)
```
zhrt123, line 758: How would `curDB->dbid != curDBId` happen?

diskquota.c
```diff
@@ -735,10 +736,48 @@ disk_quota_launcher_main(Datum main_arg)
...
+		{
+			advance_one_db = true;
+			continue;
+		}
+
+		/*
+		 * Try to start a worker to run the db if has exceeded the next_run_time
```
zhrt123, line 765:
```suggestion
		 * Try to start a worker to run the db if has exceeded the next_run_time.
```
Add a `.` to avoid confusion.

diskquota.c
fanfuxiaoran: As the comment said: we first pick up a db as curDB to run , then may fall asleep, if during the sleep time , the cur db extension is dropped by the user, and another db's extension has been created and resued the slot, then the case happens. dbArrray is in shared memory, they are slots for DiskquotaDBEntry.

fanfuxiaoran: hmm... I should add some test cases for it.

diskquota_utility.c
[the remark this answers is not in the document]
fanfuxiaoran: Fixed

diskquota.c
fanfuxiaoran: Fixed

diskquota.c
```diff
@@ -1240,27 +1279,28 @@ terminate_all_workers(void)
...
-	/* pick a db run */
-	LWLockAcquire(diskquota_locks.dblist_lock, LW_EXCLUSIVE);
-	dbEntry = next_db();
-	if (dbEntry == DiskquotaLauncherShmem->dbArrayTail) goto Failed;
-
-	dbEntry->workerId  = dq_worker->id;
-	dq_worker->dbEntry = dbEntry;
+	if (dq_worker == NULL)
+	{
+		elog(DEBUG1, "[diskquota] no free workers");
```
zhrt123, line 1293: Should we skip busy workers and search for the next worker, or use a priority queue to maintain workers?

diskquota.c
zhrt123: Got it.

diskquota.c
fanfuxiaoran: Hmm... this is a little complicated issue.

1. I will remove line 705th, as have set the nap and it's not 0, no need to check the diskquota.naptime anymore.
2. > Firstly, if the sleep time is too short, we should also set nap.tv_usec to 1, as the comment wait at least one time slice, avoid 100% CPU usage

I think when the diskquota launcher has nothing to do (no db or no free worker),  and the diskquota.nap time is 0, we need to let the diskquota launcher sleep for a while instead of polling and causing CPU busy. In other cases, there is no need to check if the diskquota.naptime is 0. ( Although there are a lot of checks  in diskquota now, but doesn't make sense). Such as if we set the diskquota.naptime to 0,  we should let it work without a break when there are dbs and enough free workers to run.

diskquota.h
```diff
@@ -155,19 +156,25 @@ struct DiskQuotaWorkerEntry
...
 	DiskquotaDBEntry *dbArray;        // size == MAX_NUM_MONITORED_DB
 	DiskquotaDBEntry *dbArrayTail;
-	int               running_workers_num;
 	volatile bool     isDynamicWorker;
 	/*
 	DiskQuotaWorkerEntry worker[diskquota_max_workers]; // the hidden memory to store WorkerEntry
 	DiskquotaDBEntry     dbentry[MAX_NUM_MONITORED_DB]; // the hidden memory for dbentry
 	*/
 } DiskquotaLauncherShmemStruct;
+extern DiskquotaLauncherShmemStruct *DiskquotaLauncherShmem;
```
higuoxing, line 177: Can we put the definition of `DiskquotaLauncherShmem` here and remove the `extern` keyword?

higuoxing (approved):
LGTM!

diskquota.h
fanfuxiaoran: Ok

PR #246, opened 2022-10-27
Fix test_rejectmap flaky test

fanfuxiaoran:
As the table has been created, its relid is a valid oid, then `check_rejectmap_by_reloid(reloid)` in `quota_check_common` will be executed and quota exceeded error will be reported with schema name or role name or tablespace name.

Why the test case can run successfully before, but sometimes fails? Because it adds the reject entry after starting inserting data into the table. If the inserting has hang up on `check_rejectmap_by_relfilenode` and then adding the reject entry, hardlimit will be triggered. Otherwise, softlimit will be triggered.

### DIFF
--- a/quotamodel.c
+++ b/quotamodel.c
@@ -1557,7 +1557,6 @@ quota_check_common(Oid reloid, RelFileNode *relfilenode)
 #ifdef FAULT_INJECTOR
 	if (SIMPLE_FAULT_INJECTOR("enable_check_quota_by_relfilenode") == FaultInjectorTypeSkip) enable_hardlimit = true;
 #endif
-	SIMPLE_FAULT_INJECTOR("check_rejectmap_by_relfilenode");
 	if (relfilenode && enable_hardlimit) return check_rejectmap_by_relfilenode(*relfilenode);
 
 	return true;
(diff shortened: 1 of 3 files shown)


quotamodel.c
```diff
@@ -1557,7 +1557,7 @@ quota_check_common(Oid reloid, RelFileNode *relfilenode)
 #ifdef FAULT_INJECTOR
 	if (SIMPLE_FAULT_INJECTOR("enable_check_quota_by_relfilenode") == FaultInjectorTypeSkip) enable_hardlimit = true;
 #endif
-
+	SIMPLE_FAULT_INJECTOR("check_rejectmap_by_relfilenode");
```
higuoxing, line 1560: We've already had this line inside `check_rejectmap_by_relfilenode()`. Why we need to add another injection point here?

higuoxing (approved):
LGTM with one question.

quotamodel.c
fanfuxiaoran: I think it should be defined outside the function, exactly, it should be defined in the caller. I will recheck it.

fanfuxiaoran: Sorry, I misunderstand it. You are right, this line is not needed, will remove it. Thanks!

PR #250, opened 2022-11-03
Fix flaky test

fanfuxiaoran:
When diskquota is in dynamic mode, it may take a long time to start a worker for a database (because the worker is busy), so some tests will fail after pg_sleep.

Add the diskquota_test extension to resolve the issue. In it, there is a UDF `diskquota_test.wait(sql test)`, we can run pass any SQL whose select result is true or false to it, it runs the SQL and waits for the select result to be true at most 10 seconds.  `diskquota_test.db_status` will return the all the databases which are monitored by diskquota and their status and epoch.  `diskquota_test.cur_db_status()` and `diskquota_test.db_status()` are just wrappers of it, which are more useful in the tests.

In the regress tests, replace the sleep with diskquota_test.wait or diskquota.wait_for_worker_new_epoch().

What's more, I have extracted the monitored_dbid_cache related data and functions into the monitored_db.c file.
Disable the regress test `test_init_table_size_table`, there is a bug making it fail, I will fix it in another pr.


tests/regress/sql/config.sql
```diff
@@ -5,7 +5,7 @@ CREATE DATABASE diskquota;
 \! gpconfig -c diskquota.naptime -v 0 --skipvalidation
 \! gpconfig -c max_worker_processes -v 20 --skipvalidation
 \! gpconfig -c diskquota.hard_limit -v "off" --skipvalidation
-\! gpconfig -c diskquota.max_workers -v 1 --skipvalidation
+\! gpconfig -c diskquota.max_workers -v 10 --skipvalidation
```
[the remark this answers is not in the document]
fanfuxiaoran: Maybe can use the epoch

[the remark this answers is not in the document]
fanfuxiaoran: I have tried the gp_inject_falut, it can help to check if there is any bgworker started for a database, but too many sql lines of it, and I think the UDF `diskquota.wait_for_worker_new_epoch()` can also help to do this work.

I have no better idea to track the bgworker behaves rather than log  to help debug now.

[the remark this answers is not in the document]
fanfuxiaoran: For most tests, it is enough.  Like the test_extension.sql, in which there are many cases to wait for  'refresh_disk_quota_model' to finish.

In some other tests, check the diskquota behaviors when the database is not ready to run, the `SELECT diskquota.wait_for_worker_new_epoch();` obviously doesn't work. But, actually, when the database is not ready to run,  it should exit and free the bgworker instead of sleeping there and holding the bgworker.  I plan to do it in the next release.  So I think it's not time to add some hooks here to track the diskquota bgworker behavers now.

monitored_db.c
```diff
@@ -0,0 +1,329 @@
...
+#include "commands/dbcommands.h"
+#include "storage/proc.h"
+#include "utils/builtins.h"
+
+PG_FUNCTION_INFO_V1(show_worker_epoch);
+PG_FUNCTION_INFO_V1(db_status);
+PG_FUNCTION_INFO_V1(wait_for_worker_new_epoch);
+
+HTAB       *monitored_dbid_cache = NULL; // Map<Oid, MonitorDBEntryStruct>
+const char *DBStatusToString[]   = {"INIT", "ERROR", "UNREADY", "PAUSED", "RUNNING", "UNKNOWN"};
```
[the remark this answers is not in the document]
fanfuxiaoran: Ok, will put them together in the diskquota.h

monitored_db.c
```diff
@@ -0,0 +1,335 @@
...
+	MonitorDBEntry entry;
+	bool           found;
+	LWLockAcquire(diskquota_locks.monitored_dbid_cache_lock, LW_SHARED);
+	{
+		entry = hash_search(monitored_dbid_cache, &dbid, HASH_FIND, &found);
+	}
+	if (found)
+	{
+		Assert(status < DB_STATUS_MAX);
+		pg_atomic_write_u32(&(entry->status), status);
```
zhrt123, line 294: Since we got the lock before, should we not atomically write status to entry->status?

fanfuxiaoran: The lock we got is a shared lock, cannot be used to change the entry->status

diskquota.c
```diff
@@ -477,6 +420,7 @@ disk_quota_worker_main(Datum main_arg)
 		{
 			break;
 		}
+		if (count == 1) update_monitordb_status(MyWorkerInfo->dbEntry->dbid, DB_UNREADY);
```
zhrt123, line 423: If we want to set `DB_UNREADY` status here and leave the loop, `count` seems to be confusing. Should we change it to `bool is_unready = false`?

fanfuxiaoran: We just set DB_UNREADY, not to leave the loop as the db is unready. Need to continue to check if the db is ready.

monitored_db.c
zhrt123: Got it.

tests/regress/diskquota_schedule
```diff
@@ -1,7 +1,7 @@
 test: config
 test: test_create_extension
 test: test_readiness_logged
-test: test_init_table_size_table
+#test: test_init_table_size_table
```
zhrt123, line 4: We can remove this change after merging https://github.com/greenplum-db/diskquota/pull/254.

diskquota.c
zhrt123: it's my mistake. But I think `is_unready` is easier to understand than `count`.

diskquota.c
```diff
@@ -399,11 +399,11 @@ disk_quota_worker_main(Datum main_arg)
 	 */
 	init_ps_display("bgworker:", "[diskquota]", dbname, "");
 
-	int count = 0;
+	/* suppose the database is ready, if not, then set it to false */
+	bool is_ready = true;
```
zhrt123, line 403: it should be `is_unready` instead of `is_ready`. If the database is ready, it should leave this loop by `check_diskquota_state_is_ready()`. So we should set `is_unready = false` initially and set `is_unready = true` after `update_monitordb_status()`.

diskquota.c
```diff
@@ -475,8 +418,14 @@ disk_quota_worker_main(Datum main_arg)
...
 		if (check_diskquota_state_is_ready())
 		{
+			is_ready = true;
 			break;
 		}
+		if (is_ready)
+		{
+			update_monitordb_status(MyWorkerInfo->dbEntry->dbid, DB_UNREADY);
+			is_ready = false;
+		}
```
zhrt123, line 428:
```suggestion
		if (is_unready == false)
		{
			update_monitordb_status(MyWorkerInfo->dbEntry->dbid, DB_UNREADY);
			is_unready = true;
		}
```

diskquota.c
```diff
@@ -475,8 +418,14 @@ disk_quota_worker_main(Datum main_arg)
 		 */
 		if (check_diskquota_state_is_ready())
 		{
+			is_ready = true;
```
zhrt123, line 421:
```suggestion
```

diskquota.c
```diff
@@ -458,6 +399,8 @@ disk_quota_worker_main(Datum main_arg)
 	 */
 	init_ps_display("bgworker:", "[diskquota]", dbname, "");
 
+	/* suppose the database is ready, if not, then set it to false */
+	bool is_ready = true;
```
zhrt123, line 403:
```suggestion
	bool is_unready = false;
```

diskquota.c
fanfuxiaoran: I think supposing the database is ready to run before checking is better. Then if it is unready, we set it to DB_UNREADY.

zhrt123: In that case, we need to set `is_ready = false` initially, correct?

fanfuxiaoran: In the MonitorDBStatus, there is no one status called READY. There are INIT, UNREADY, and RUNNING.   When diskquota is ready to run, it will turn from INIT to RUNNING, otherwise, it will turn INIT to UNREADY. So we suppose that it is ready in begin.

zhrt123: Got it. I'm just confused by `check_diskquota_state_is_ready()`.

zhrt123 (approved):

PR #257, opened 2022-11-18
Change the max value of diskquota.max_workers to 20

fanfuxiaoran:
If we set the diskquota.max_workers max value to be max_worker_processes, when max_worker_processes is less than 10 and we set diskquota.max_workers value more than max_worker_processes, the cluster will crash.

Set the max value to 20, when the max_worker_processes is less than the diskquota.max_workers, diskquota can work, only some parts of the databases can not be monitored as diskquota can not start bgworkers for them.

### DIFF
--- a/diskquota.c
+++ b/diskquota.c
@@ -595,7 +595,7 @@ disk_quota_launcher_main(Datum main_arg)
 			if (curDB != NULL)
 			{
 				curDBId = curDB->dbid;
-				elog(DEBUG1, "[diskquota] next db to run:%d", curDB->id);
+				elog(DEBUG1, "[diskquota] next db to run:%u", curDB->dbid);
 			}
 			else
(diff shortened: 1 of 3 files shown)


beeender (approved):

diskquota.c
```diff
@@ -595,7 +595,7 @@ disk_quota_launcher_main(Datum main_arg)
 			if (curDB != NULL)
 			{
 				curDBId = curDB->dbid;
-				elog(DEBUG1, "[diskquota] next db to run:%d", curDB->id);
+				elog(DEBUG1, "[diskquota] next db to run:%u", curDB->dbid);
```
zhrt123, line 598:
```suggestion
				elog(DEBUG1, "[diskquota] next db to run:%u", curDBId);
```

zhrt123 (approved):

PR #265, opened 2022-11-28
WIP:fix deadlock when droping extension

fanfuxiaoran:


fanfuxiaoran:
Seems there is no need to fix it.

PR #266, opened 2022-11-29
Change `diskquota.max_active_tables` default value to reduce memory cost

zhrt123:
Change the default value of `diskquota.max_active_tables` from 1M to 300K, and the memory usage relevant is reduced from 300MB to 90MB.

### DIFF
--- a/diskquota.c
+++ b/diskquota.c
@@ -280,7 +280,7 @@ define_guc_variables(void)
 	                        2, min_naptime, INT_MAX, PGC_SIGHUP, 0, NULL, NULL, NULL);
 
 	DefineCustomIntVariable("diskquota.max_active_tables", "Max number of active tables monitored by disk-quota.", NULL,
-	                        &diskquota_max_active_tables, 1 * 1024 * 1024, 1, INT_MAX, PGC_SIGHUP, 0, NULL, NULL, NULL);
+	                        &diskquota_max_active_tables, 300 * 1024, 1, INT_MAX, PGC_POSTMASTER, 0, NULL, NULL, NULL);
 
 	DefineCustomIntVariable("diskquota.worker_timeout", "Duration between each check (in seconds).", NULL,
 	                        &diskquota_worker_timeout, 60, 1, INT_MAX, PGC_SIGHUP, 0, NULL, NULL, NULL);
--- a/tests/isolation2/expected/config.out
+++ b/tests/isolation2/expected/config.out
@@ -21,7 +21,7 @@
 1: SHOW diskquota.max_active_tables;
  diskquota.max_active_tables 
 -----------------------------
- 1048576                     
+ 307200                      
 (1 row)
 1: SHOW diskquota.worker_timeout;
  diskquota.worker_timeout 
--- a/tests/regress/expected/config.out
+++ b/tests/regress/expected/config.out
@@ -11,7 +11,7 @@ SHOW diskquota.naptime;
 SHOW diskquota.max_active_tables;
  diskquota.max_active_tables 
 -----------------------------
- 1048576
+ 307200
 (1 row)
 
 SHOW diskquota.worker_timeout;


diskquota.c
```diff
@@ -280,7 +281,7 @@ define_guc_variables(void)
 	                        2, min_naptime, INT_MAX, PGC_SIGHUP, 0, NULL, NULL, NULL);
 
 	DefineCustomIntVariable("diskquota.max_active_tables", "Max number of active tables monitored by disk-quota.", NULL,
-	                        &diskquota_max_active_tables, 1 * 1024 * 1024, 1, INT_MAX, PGC_SIGHUP, 0, NULL, NULL, NULL);
+	                        &diskquota_max_active_tables, 200 * 1024, 1, INT_MAX, PGC_SIGHUP, 0, NULL, NULL, NULL);
```
beeender, line 284: I think we agree to `300,000` as the default value?

beeender: I think `POSTMASTER` should be used instead of `PGC_SIGHUP`. To change this GUC, user has to restart the cluster. `gpstop -u` won't recreate shared memory?

diskquota.c
```diff
@@ -290,6 +291,9 @@ define_guc_variables(void)
 	        "diskquota.max_workers",
 	        "Max number of backgroud workers to run diskquota extension, should be less than max_worker_processes.",
 	        NULL, &diskquota_max_workers, 10, 1, 20, PGC_POSTMASTER, 0, NULL, NULL, NULL);
+	DefineCustomIntVariable("diskquota.max_tables_all_segments", "Max number of tables on all segments.", NULL,
+	                        &diskquota_max_tables_all_segments, 2 * 1024 * 1024, 1, INT_MAX, PGC_SIGHUP, 0, NULL, NULL,
```
beeender, line 295: ditto

quotamodel.c
```diff
@@ -452,10 +453,9 @@ init_lwlocks(void)
 static Size
 diskquota_worker_shmem_size()
 {
-	Size size;
-	size = hash_estimate_size(MAX_TABLES, sizeof(TableSizeEntry));
-	size = add_size(size, hash_estimate_size(MAX_LOCAL_DISK_QUOTA_REJECT_ENTRIES, sizeof(LocalRejectMapEntry)));
-	size = add_size(size, hash_estimate_size(1024L, sizeof(struct QuotaMapEntry)) * NUM_QUOTA_TYPES);
+	Size size = 0;
```
beeender, line 456: `size = hash_estimate_size(MAX_TABLES, sizeof(TableSizeEntry));`

hmm, is this a bug before? We don't need `TableSizeTable` on QEs, do we?

diskquota.c
zhrt123: I think `gpstop -u` will re-allocate shared memory, I'll try it.

zhrt123: > I think we agree to 300,000 as the default value?

I've changed it.

quotamodel.c
zhrt123: No, this is used by bgworkers, not QEs.

diskquota.c
```diff
@@ -290,6 +291,9 @@ define_guc_variables(void)
 	        "diskquota.max_workers",
 	        "Max number of backgroud workers to run diskquota extension, should be less than max_worker_processes.",
 	        NULL, &diskquota_max_workers, 10, 1, 20, PGC_POSTMASTER, 0, NULL, NULL, NULL);
+	DefineCustomIntVariable("diskquota.max_tables_all_segments", "Max number of tables on all segments.", NULL,
```
radarwave, line 294: Should be max_tables_on_all_segments?

And how about using the name: diskquota.max_table_shards_on_segments

beeender (approved):
Also need to discuss this with @radarwave about the GUC name before merging

diskquota.c
zhrt123: Good idea.

beeender (approved):

fanfuxiaoran (approved):

radarwave (approved):

PR #267, opened 2022-11-29
Move llocal_disk_quota_reject_map from shared memory to local memory

fanfuxiaoran:
Move local_disk_quota_reject_map from shared memory to local memory
Before this change, local_disk_quota_reject_map is saved in shared memory. In each loop, (it saves all the reject entries of the last loop), diskquota marks the segexceeded and isexceeded of each item.  When flushing it to disk_quota_reject_map(global reject map), firstly flush all the exceeded ones, for unexceeded items, remove them in both map.
After moving it to local memory, in each flush, firstly remove all reject items of this database in disk_quota_reject_map, then flush all items in local_disk_quota_reject_map to it and clean up the local_disk_quota_reject_map .

### DIFF
--- a/quotamodel.c
+++ b/quotamodel.c
@@ -1115,6 +1115,8 @@ flush_local_reject_map(void)
 	hash_seq_init(&iter, local_disk_quota_reject_map);
 	while ((localrejectentry = hash_seq_search(&iter)) != NULL)
 	{
+		/* Ethier the totalsize of table exceed the quota,or the tablesize on segment exceed the segment quota, the
+		 * isexceed is set to be true*/
 		if (localrejectentry->isexceeded)
 		{
 			rejectentry = (GlobalRejectMapEntry *)hash_search(disk_quota_reject_map, (void *)&localrejectentry->keyitem,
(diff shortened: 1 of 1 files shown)


quotamodel.c
```diff
@@ -1123,9 +1102,21 @@ flush_local_reject_map(void)
...
 
+	hash_seq_init(&iter, disk_quota_reject_map);
+
+	while ((rejectentry = hash_seq_search(&iter)) != NULL)
+	{
+		if (rejectentry->keyitem.databaseoid == MyDatabaseId)
+		{
+			(void)hash_search(disk_quota_reject_map, (void *)&rejectentry->keyitem, HASH_REMOVE, NULL);
+		}
+	}
```
zhrt123, line 1113: We need only to use SharedLock to lock this loop, isn't it?

fanfuxiaoran: > We need only to use SharedLock to lock this loop, isn't it?

No, we need LW_EXCLUSIVE lock, as local_active_table_stat_map is global map and we remove entry in it here.

fanfuxiaoran:
Diskquota will be refactored in 3.0, so just close it.

PR #264, opened 2022-11-28
Reduce memory usage.

zhrt123:
Refactor the structure of `TableSizeEntry` to reduce memory usage.

Previously, the size of each table in each segment should be maintained in `TableSizeEntry`, which wastes lots of memory. In this PR, we refactor the `TableSizeEntry` to:
```
struct TableSizeEntry
{
	Oid    reloid;
	int    segid;
	Oid    tablespaceoid;
	Oid    namespaceoid;
	Oid    owneroid;
	uint32 flag;
	int64 totalsize[SEGMENT_SIZE_ARRAY_LENGTH];
};
```

In this way, we can maintain multiple sizes in one `TableSizeEntry` and efficiently save memory usage.
- For 50 segments: reduced by 65%.
- For 100 segments: reduced by 82.5%.
- For 101 segments: reduced by 65.3%.
- For 1000 segments: reduced by 82.5%.

PR #268, opened 2022-11-29
Fix table entry key

fanfuxiaoran:
There is a bug: removing TableSizeEntry from table_size_map by oid.
Actually the hash map key is TableKeyEntry. Fix it.


zhrt123 (approved):

beeender (commented):
nice catch!
I will check this again after it rebased (#264 merged)

zhrt123 (approved):

beeender (approved):

PR #271, opened 2022-12-02
Add a GUC `diskquota.max_table_segments`.

zhrt123:
Use `diskquota.max_table_segments` to define the max number of table segments in the cluster. The value equal `(segment_number + 1) * max_table_number`.


radarwave:
Let's set the default value of diskquota.max_table_segments to 10 * 1024 * 1024, the memory usage is acceptable and can support more tables.

quotamodel.c
```diff
@@ -950,6 +950,19 @@ calculate_table_disk_usage(bool is_init, HTAB *local_active_table_stat_map)
 			tsentry = (TableSizeEntry *)hash_search(table_size_map, &key, HASH_ENTER, &table_size_map_found);
 			if (!table_size_map_found)
 			{
+				int counter = pg_atomic_add_fetch_u32(diskquota_table_size_entry_num, 1);
+				if (counter > MAX_NUM_TABLE_SIZE_ENTRIES)
+				{
+					hash_search(table_size_map, &key, HASH_REMOVE, NULL);
+					if ((counter - MAX_NUM_TABLE_SIZE_ENTRIES) % 100 == 1)
+					{
+						ereport(WARNING, (errmsg("[diskquota] the number of tables exceeds the limit, please increate "
```
zhrt123, line 959:
```suggestion
						ereport(WARNING, (errmsg("[diskquota] the number of tables exceeds the limit, please increase "
```

radarwave (approved):

quotamodel.c
```diff
@@ -510,6 +506,7 @@ DiskQuotaShmemSize(void)
 	if (IS_QUERY_DISPATCHER())
 	{
 		size = add_size(size, diskquota_launcher_shmem_size());
+		size = add_size(size, hash_estimate_size(MAX_NUM_TABLE_SIZE_ENTRIES, sizeof(TableSizeEntry)));
```
beeender, line 509: should `diskquota_table_size_entry_num` be added to the size?

quotamodel.c
```diff
@@ -950,6 +950,19 @@ calculate_table_disk_usage(bool is_init, HTAB *local_active_table_stat_map)
 			tsentry = (TableSizeEntry *)hash_search(table_size_map, &key, HASH_ENTER, &table_size_map_found);
 			if (!table_size_map_found)
 			{
+				int counter = pg_atomic_add_fetch_u32(diskquota_table_size_entry_num, 1);
+				if (counter > MAX_NUM_TABLE_SIZE_ENTRIES)
+				{
+					hash_search(table_size_map, &key, HASH_REMOVE, NULL);
+					if ((counter - MAX_NUM_TABLE_SIZE_ENTRIES) % 100 == 1)
+					{
+						ereport(WARNING, (errmsg("[diskquota] the number of tables exceeds the limit, please increase "
```
beeender, line 959: log the `counter` as well. easier for debugging

quotamodel.c
```diff
@@ -950,6 +950,19 @@ calculate_table_disk_usage(bool is_init, HTAB *local_active_table_stat_map)
 			tsentry = (TableSizeEntry *)hash_search(table_size_map, &key, HASH_ENTER, &table_size_map_found);
 			if (!table_size_map_found)
 			{
+				int counter = pg_atomic_add_fetch_u32(diskquota_table_size_entry_num, 1);
+				if (counter > MAX_NUM_TABLE_SIZE_ENTRIES)
+				{
+					hash_search(table_size_map, &key, HASH_REMOVE, NULL);
```
beeender, line 956: I don't think we should remove the entry. even the entry is removed, the share-mem cannot be used by others.
I suggest just leave it there, log a warning, remove the line 964 continue.

quotamodel.c
```diff
@@ -950,6 +950,19 @@ calculate_table_disk_usage(bool is_init, HTAB *local_active_table_stat_map)
...
+					if ((counter - MAX_NUM_TABLE_SIZE_ENTRIES) % 100 == 1)
```
beeender, line 957: `- MAX_NUM_TABLE_SIZE_ENTRIES` is not necessary. `counter % 100 == 1` is good enough.

quotamodel.c
zhrt123: No

quotamodel.c
zhrt123: If we don't remove it, the memory will overflow after many tables are created.

beeender: even we remove it, the sh-mem slot cannot be recycled I think?

fanfuxiaoran:
In my opinion, we may not need the diskquota_table_size_entry_num to control the use of shared memory. Or we need more consideration about it.   @zhrt123 , could you extract it from this pr, and create a new pr for it?

quotamodel.c
```diff
@@ -950,6 +950,19 @@ calculate_table_disk_usage(bool is_init, HTAB *local_active_table_stat_map)
...
+					{
+						ereport(WARNING, (errmsg("[diskquota] the number of tables exceeds the limit, please increate "
+						                         "the GUC value for diskquota.max_table_segments. Current "
+						                         "diskquota.max_table_segments value: %d",
+						                         diskquota_max_table_segments)));
+					}
+					continue;
```
zhrt123, line 964:
```suggestion
					break;
```

diskquota.c
```diff
@@ -83,6 +84,9 @@ static DiskQuotaWorkerEntry *volatile MyWorkerInfo = NULL;
 // how many database diskquota are monitoring on
 static int num_db = 0;
 
+/* how many TableSizeEntry are maintained in table_size_map */
```
fanfuxiaoran, line 87:
```suggestion
/* how many TableSizeEntry are maintained in all the table_size_map in shared memory*/
```

diskquota.h
```diff
@@ -30,6 +30,13 @@
 
 #include <signal.h>
 
+/* init number of TableSizeEntry in table_size_map */
+#define INIT_NUM_TABLE_SIZE_ENTRIES 128
+/* max number of TableSizeEntry in table_size_map */
+#define MAX_NUM_TABLE_SIZE_ENTRIES (diskquota_max_table_segments / SEGMENT_SIZE_ARRAY_LENGTH)
+/* length of segment size array in TableSizeEntry */
+#define SEGMENT_SIZE_ARRAY_LENGTH 100
```
fanfuxiaoran, line 38: Move them to quotamodel.c?

quotamodel.c
```diff
@@ -84,7 +79,9 @@ typedef struct RejectMapEntry       RejectMapEntry;
 typedef struct GlobalRejectMapEntry GlobalRejectMapEntry;
 typedef struct LocalRejectMapEntry  LocalRejectMapEntry;
 
-int SEGCOUNT = 0;
+int                      SEGCOUNT = 0;
+extern int               diskquota_max_table_segments;
+extern pg_atomic_uint32 *diskquota_table_size_entry_num;
```
fanfuxiaoran, line 84: Better to extern them in diskquota.h

beeender:
> In my opinion, we may not need the diskquota_table_size_entry_num to control the use of shared memory. Or we need more consideration about it. @zhrt123 , could you extract it from this pr, and create a new pr for it?

We had an offline discussion about this. Diskqutoa should NOT use too much memory than it planed to use, to avoid interrupt other database functionalities. When there are more tables than expected, diskquota may not be able to calculate the right quota size. But that is accept IMO. User could find the warning in the log and enlarge the `diskquota_max_table_segments`.

quotamodel.c
```diff
@@ -510,6 +506,7 @@ DiskQuotaShmemSize(void)
 	if (IS_QUERY_DISPATCHER())
 	{
 		size = add_size(size, diskquota_launcher_shmem_size());
```
fanfuxiaoran, line 508: Have you add the  shared memory size for `diskquota_table_size_entry_num`

quotamodel.c
fanfuxiaoran, line 509: Move it to diskquota_worker_shmem_size, and multiply the `MAX_NUM_MONITORED_DB` the function `MAX_NUM_MONITORED_DB`.

diskquota.h
```diff
@@ -30,6 +30,13 @@
 
 #include <signal.h>
 
+/* init number of TableSizeEntry in table_size_map */
+#define INIT_NUM_TABLE_SIZE_ENTRIES 128
```
fanfuxiaoran, line 34: Is this too small? And move it to quotamodel.c?

quotamodel.c
fanfuxiaoran: Why? as it is in the shared memory?

quotamodel.c
fanfuxiaoran: A comment should be added here, a confusing line. And I think this may lead to too log .

quotamodel.c
zhrt123: `MAX_NUM_TABLE_SIZE_ENTRIES` is the total number of entries in all databases.

quotamodel.c
zhrt123:
```suggestion
		size = add_size(size, diskquota_launcher_shmem_size());
		size = add_size(size, sizeof(pg_atomic_uint32));
```

quotamodel.c
```diff
@@ -953,10 +954,22 @@ calculate_table_disk_usage(bool is_init, HTAB *local_active_table_stat_map)
 			tsentry = (TableSizeEntry *)hash_search(table_size_map, &key, HASH_ENTER, &table_size_map_found);
 			if (!table_size_map_found)
 			{
+				int counter = pg_atomic_add_fetch_u32(diskquota_table_size_entry_num, 1);
```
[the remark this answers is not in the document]
zhrt123: It does not influence too much memory. However if we check the counter first, we don't know whether we will find the entry or not.

quotamodel.c
zhrt123: But we create hashtable with `max_size = MAX_NUM_TABLE_SIZE_ENTRIES` for each database's table_size_map, if we allocate it with `MAX_NUM_TABLE_SIZE_ENTRIES / 50`, the size is also incorrect.

fanfuxiaoran: > But we create hashtable with max_size = MAX_NUM_TABLE_SIZE_ENTRIES for each database's table_size_map

diskquota.h
radarwave: I think both options have a tiny impact, let's keep it as the current value. We can make the change when we have good reasons.

beeender:
@fanfuxiaoran @zhrt123 I am merging this to trigger the release pipeline. Please help to check this again afterward. We can always revert this. Thanks!

PR #272, opened 2022-12-02
Fix regression caused by #264

beeender:
The #264 caused some segment ratio tests fail. The entry's relevant
fields need to be set at the end of the iteration. Otherwise, only the
first seg will pass the condition check.

PR #273, opened 2022-12-05
Fix bug for #264.

zhrt123:
The #264 caused some segment ratio tests fail, we can only update the quota for the first segment's size in TableSizeEntry. This PR fixed the bug.

### DIFF
--- a/quotamodel.c
+++ b/quotamodel.c
@@ -945,7 +945,7 @@ calculate_table_disk_usage(bool is_init, HTAB *local_active_table_stat_map)
 		 * and the content id is continuous, so it's safe to use SEGCOUNT
 		 * to get segid.
 		 */
-		for (int i = -1; i < SEGCOUNT; i++)
+		for (int i = -1; i < SEGCOUNT; i += SEGMENT_SIZE_ARRAY_LENGTH)
 		{
 			key.reloid = relOid;
 			key.id     = TableSizeEntryId(i);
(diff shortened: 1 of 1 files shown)


beeender (approved):

quotamodel.c
```diff
@@ -960,82 +960,94 @@ calculate_table_disk_usage(bool is_init, HTAB *local_active_table_stat_map)
...
 
 			if (tsentry->tablespaceoid != reltablespace)
 			{
-				transfer_table_for_quota(TableSizeEntryGetSize(tsentry, i), NAMESPACE_TABLESPACE_QUOTA,
-				                         (Oid[]){tsentry->namespaceoid, tsentry->tablespaceoid},
-				                         (Oid[]){tsentry->namespaceoid, reltablespace}, i);
-				transfer_table_for_quota(TableSizeEntryGetSize(tsentry, i), ROLE_TABLESPACE_QUOTA,
-				                         (Oid[]){tsentry->owneroid, tsentry->tablespaceoid},
-				                         (Oid[]){tsentry->owneroid, reltablespace}, i);
+				for (int cur_seg = seg_st; cur_seg < seg_ed; cur_seg++)
```
fanfuxiaoran, line 1042: Too ugly to use the for loop here, and. the performance is poor.

beeender:
#272 was merged to replace this PR.

PR #278, opened 2022-12-08
Add a comment about quota_check_common

fanfuxiaoran:


fanfuxiaoran:
This pr is not needed any more

PR #279, opened 2022-12-08
Fix bug: rejectmap entries should not be removed by other databases.

zhrt123:
Previously, refresh_rejectmap() cleared all entries in rejectmap, including other databases' entries, which causes hardlimit can not to work correctly. This commit fixed the bug.

Meanwhile, this PR fixed a bug: soft-limit rejectmap entries should not be added into `disk_quota_reject_map` on segments.

### DIFF
--- a/quotamodel.c
+++ b/quotamodel.c
@@ -2040,7 +2040,10 @@ refresh_rejectmap(PG_FUNCTION_ARGS)
 	/* Clear rejectmap entries. */
 	hash_seq_init(&hash_seq, disk_quota_reject_map);
 	while ((rejectmapentry = hash_seq_search(&hash_seq)) != NULL)
+	{
+		if (rejectmapentry->keyitem.relfilenode.dbNode != MyDatabaseId) continue;
 		hash_search(disk_quota_reject_map, &rejectmapentry->keyitem, HASH_REMOVE, NULL);
+	}
 
 	/* Flush the content of local_rejectmap to the global rejectmap. */
 	hash_seq_init(&hash_seq, local_rejectmap);
--- a/quotamodel.c
+++ b/quotamodel.c
@@ -811,7 +811,8 @@ refresh_disk_quota_usage(bool is_init)
 		flush_to_table_size();
 		/* copy local reject map back to shared reject map */
 		bool reject_map_changed = flush_local_reject_map();
-		/* Dispatch rejectmap entries to segments to perform hard-limit.
+		/*
+		 * Dispatch rejectmap entries to segments to perform hard-limit.
 		 * If the bgworker is in init mode, the rejectmap should be refreshed anyway.
 		 * Otherwise, only when the rejectmap is changed or the active_table_list is
 		 * not empty the rejectmap should be dispatched to segments.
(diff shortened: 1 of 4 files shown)


fanfuxiaoran (approved):

quotamodel.c
```diff
@@ -2038,24 +2044,32 @@ refresh_rejectmap(PG_FUNCTION_ARGS)
 	LWLockAcquire(diskquota_locks.reject_map_lock, LW_EXCLUSIVE);
 
 	/* Clear rejectmap entries. */
+	/* Pass two empty arrays to clear the reject map. */
+	force_clean = (reject_array_count == 0) && (active_array_count == 0);
```
fanfuxiaoran, line 2048: Even if force_clean is true, I think we should only delete the object that belong to the current db.   When all the db workers work well, and another diskquota extension is created, it will do the init things, this will lead to all the rejected map entries being cleaned up.

beeender: `force_clean` is to satisfy the assumption of isolation2 tests `test_rejectmap`. If there is still concerns, maybe we can use `NULL` values for clear?

quotamodel.c
```diff
@@ -2038,24 +2043,31 @@ refresh_rejectmap(PG_FUNCTION_ARGS)
 	LWLockAcquire(diskquota_locks.reject_map_lock, LW_EXCLUSIVE);
 
 	/* Clear rejectmap entries. */
+	/* Pass two empty arrays to clear the reject map. */
```
zhrt123, line 2046:
```suggestion
```

quotamodel.c
```diff
@@ -2038,24 +2043,31 @@ refresh_rejectmap(PG_FUNCTION_ARGS)
...
 		bool                  found;
 		GlobalRejectMapEntry *new_entry;
-		new_entry = hash_search(disk_quota_reject_map, &rejectmapentry->keyitem, HASH_ENTER_NULL, &found);
+
 		/*
-		 * We don't perform soft-limit on segment servers, so we don't flush the
-		 * rejectmap entry with a valid targetoid to the global rejectmap on segment
-		 * servers.
+		 * Skip soft limit reject entry We don't perform soft-limit on segment servers, so we don't flush the
+		 * rejectmap entry with a valid targetoid to the global rejectmap on segment servers.
```
zhrt123, line 2065:
```suggestion
		 * Skip soft limit reject entry. We don't perform soft-limit on segment servers, so we don't flush the
		 * rejectmap entry with a valid targetoid to the global rejectmap on segment servers.
```

quotamodel.c
```diff
@@ -811,8 +811,12 @@ refresh_disk_quota_usage(bool is_init)
 		flush_to_table_size();
 		/* copy local reject map back to shared reject map */
 		bool reject_map_changed = flush_local_reject_map();
-		/* Dispatch rejectmap entries to segments to perform hard-limit. */
-		if (diskquota_hardlimit && (reject_map_changed || hasActiveTable))
+		/* Dispatch rejectmap entries to segments to perform hard-limit.
```
zhrt123, line 814:
```suggestion
		/* 
		 * Dispatch rejectmap entries to segments to perform hard-limit.
```

fanfuxiaoran (approved):

beeender (approved):

PR #280, opened 2022-12-09
Fix diskquota worker schedule bug

fanfuxiaoran:
When a database's diskquota bgworker is a killed and the db is droped, diskquota scheduler can not work properly. The cause is: if the scheduler failed to start a bgworker for a database, it will try it again and again forever.

Fixing it by: firstly check if the database is still valid, if not, skip it in the next_db routine. Secondly, limit the times of starting a bgworker for a database, if has tried 3 times, will skip it and pick the next one.


diskquota.c
```diff
@@ -1336,6 +1342,21 @@ is_valid_dbid(Oid dbid)
...
 
+/* SearchSysCache should be run in a transaction */
+static bool
+check_valid_dbid(Oid dbid)
+{
+	bool          ret;
+	MemoryContext old_ctx;
+	StartTransactionCommand();
+	(void)GetTransactionSnapshot();
+	old_ctx = MemoryContextSwitchTo(TopMemoryContext);
```
beeender, line 1353: Why switching memory context is necessary?

diskquota.c
```diff
@@ -1336,6 +1342,21 @@ is_valid_dbid(Oid dbid)
 	return true;
 }
 
+/* SearchSysCache should be run in a transaction */
+static bool
+check_valid_dbid(Oid dbid)
```
beeender, line 1347: merge `is_valid_dbid()` and `check_valid_dbid()` into one function like:
```
check_valid_dbid(Oid dbid, bool need_transaction)
```
looks better IMO.

Minor, I can accept the current one

tests/regress/sql/test_worker_schedule_exception.sql
```diff
@@ -0,0 +1,38 @@
+-- start_ignore
+\! gpconfig -c diskquota.max_workers -v 10;
+\! gpconfig -c diskquota.naptime -v 4;
+\! gpstop -arf;
+\c
```
beeender, line 5: this is not needed

tests/regress/sql/test_worker_schedule.sql
```diff
@@ -2,6 +2,7 @@
 \c
 
 DROP DATABASE IF EXISTS t1;
+DROP DATABASE IF EXISTS t2;
```
beeender, line 5: I think you forget to commit the out file

tests/regress/sql/test_worker_schedule_exception.sql
```diff
@@ -0,0 +1,38 @@
...
+SELECT diskquota.wait_for_worker_new_epoch();
+
+\c t2
+SELECT diskquota.pause();
+SELECT diskquota.wait_for_worker_new_epoch();
+DROP EXTENSION diskquota;
+
+--start_ignore
+\c contrib_regression
+DROP DATABASE t2;
```
beeender, line 34: `DROP` statement should not be ignored.

diskquota.c
fanfuxiaoran: Oh , switching memory is not needed here, will delete it.

tests/regress/sql/test_worker_schedule_exception.sql
fanfuxiaoran: I need the diskquota to run in static mode, then I can kill the bgworker. And the diskquota.naptime should be a little longer, as if it is too small, the diskquota launcher runs so fastly that it will start another bgworker for it before dropping the database.

tests/regress/sql/test_worker_schedule.sql
fanfuxiaoran: As it is included in the ignore block, so I didn't add it. Anyway, better to add it.

tests/regress/sql/test_worker_schedule_exception.sql
fanfuxiaoran: 👌

diskquota.c
fanfuxiaoran: Yeah, this should be optimized, but I need to do more investigation about it and do that then.

tests/regress/sql/test_worker_schedule_exception.sql
beeender: `\c` is needed to reconnect because of `gpstop -arf` :+1:

beeender (approved):

diskquota.c
```diff
@@ -1336,6 +1342,21 @@ is_valid_dbid(Oid dbid)
...
+/*
+ * SearchSysCache should be run in a transaction
+ * TODO: should merge it with is_valid_dbid?
+ */
+static bool
+check_valid_dbid(Oid dbid)
+{
+	bool ret;
+	StartTransactionCommand();
+	(void)GetTransactionSnapshot();
```
beeender, line 1354: Is `GetTransactionSnapshot()` necessary?

tests/regress/sql/test_worker_schedule_exception.sql
```diff
@@ -0,0 +1,38 @@
+-- start_ignore
+\! gpconfig -c diskquota.max_workers -v 10;
+\! gpconfig -c diskquota.naptime -v 4;
+\! gpstop -arf;
+\c
+DROP DATABASE IF EXISTS t1;
+DROP DATABASE IF EXISTS t2;
```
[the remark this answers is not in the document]
fanfuxiaoran:
```
1. create diskquota extension for database t1;
2. kill bgworker of database t1
3. drop database t1
4. create diskquota extension for database t2
5. check if bgworker for t2 works
```

diskquota.c
zhrt123, line 1354: Should we push and pop this Snapshot? And to avoid multiple calls, I think it is better to start a transaction in next_db() instead of check_valid_dbid().

diskquota.c
```diff
@@ -732,6 +735,9 @@ disk_quota_launcher_main(Datum main_arg)
 			advance_one_db = ret;
 			/* has exceeded the next_run_time of current db */
 			timeout = true;
+			/* only try to start bgworker for a database at most 3 times */
+			try_times++;
+			if (try_times >= 3) advance_one_db = true;
```
zhrt123, line 740: I think we should not pick another database when `try_times >= 3`. Because when there is no free worker, start_worker() will return false, and we also can not start a worker for another database.

fanfuxiaoran: There are several reasons for the failure of starting a bgworker for a database, such as no free worker, or the database is not a valid database, or forking a process failed.

Your concern is reasonable.  The launcher can be disturbed by a signal that a bgworker is available, a bgworker is created for a database, or an extension is dropped or created.  3 is too little, I think 10 or more times is better.


> The cause is: if the scheduler failed to start a bgworker for a database, it will try it again and again forever

As the comment said, if we don't use the try times in the scheduler, diskquota will stop working in some situations, this is a disaster.

diskquota.c
fanfuxiaoran: Checked, not necessary, have removed it. And deleted the check_valid_dbid function. Added StartTransactionCommand in next_db function.

zhrt123 (approved):

beeender:
CI failure should not be related to this change. Strangely, the `sshd` doesn't exist anymore in the job container. Thus the mirror switching failed.

Merge this.

PR #284, opened 2023-01-11
Fix on gpdb7

fanfuxiaoran:
1) Fix some compile issue, especially relstorage has been removed
on gpdb7. Using relam to get the relation's storage type.

2) Disabled some tests due to AO/CO issue: As something changs about
   AO/CO table, the size of them are changed, so disable some related
   tests. Will fix them and bring them back later.

3) Upgrade to diskquota 2.2
   Add attribute relam to type relation_cache_detail and add a param to
   function relation_size_local.


upgrade_test/CMakeLists.txt
```diff
@@ -52,10 +57,6 @@ endforeach()
 
 # if DDL file modified, insure the last release file passed in
 if(DISKQUOTA_DDL_CHANGE_CHECK AND DISKQUOTA_DDL_MODIFIED AND NOT DEFINED DISKQUOTA_LAST_RELEASE_PATH)
-  message(
```
beeender, line 53: this should not be removed.

fanfuxiaoran:
The similar pr 285 has been merged.

PR #294, opened 2023-02-02
Fix flaky test

zhrt123:
- Fix flaky test test_ctas_before_set_quota. pg_type will be an active table
after `CREATE TABLE`. It does not affect the function of diskquota but
makes the test results unstable. In fact, we do not care about the table
size of the system catalog table. So we simply skip the active table oid
of these tables.

- Fix test_vacuum/test_truncate. gp_wait_until_triggered_fault
should be called after gp_inject_fault_infinite with suspend flag.

Co-authored-by: Xing Guo <higuoxing@gmail.com>
Co-authored-by: Xiaoran Wang <wxiaoran@vmware.com>

### DIFF
--- a/gp_activetable.c
+++ b/gp_activetable.c
@@ -805,7 +805,12 @@ get_active_tables_oid(void)
 		rnode.spcNode = active_table_file_entry->tablespaceoid;
 		relOid        = get_relid_by_relfilenode(rnode);
 
-		if (relOid != InvalidOid)
+		/* skip system catalog tables */
+		if (relOid < FirstNormalObjectId)
+		{
+			hash_search(local_active_table_file_map, active_table_file_entry, HASH_REMOVE, NULL);
+		}
+		else if (relOid != InvalidOid)
 		{
 			prelid             = get_primary_table_oid(relOid, true);
 			active_table_entry = hash_search(local_active_table_stats_map, &prelid, HASH_ENTER, &found);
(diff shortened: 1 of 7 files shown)


linwen:
Why run wait_for_worker_new_epoch() twice can avoid falky failure?

zhrt123:
> Why run wait_for_worker_new_epoch() twice can avoid falky failure?

It just works accidentally. The reason and solution for this flaky test is as the following:
pg_type will be an active table after `CREATE TABLE`, which makes test results unstable. In fact, we do not care about the table size of the system catalog table, so we just skip the active table oid of these tables.

gp_activetable.c
```diff
@@ -805,7 +805,12 @@ get_active_tables_oid(void)
 		rnode.spcNode = active_table_file_entry->tablespaceoid;
 		relOid        = get_relid_by_relfilenode(rnode);
 
-		if (relOid != InvalidOid)
+		/* skip system catalog tables */
+		if (relOid < FirstNormalObjectId)
+		{
+			hash_search(local_active_table_file_map, active_table_file_entry, HASH_REMOVE, NULL);
+		}
+		else if (relOid != InvalidOid)
```
fanfuxiaoran, line 813: InvalidOId is included in condition `reloid < FirstNormalObjectOid`

zhrt123: You are right, but I think it is good to show only ValidOid can enter this procedure. (To avoid InvalidOid > FirstNormalObjectId in the future too.)

fanfuxiaoran (approved):

PR #301, opened 2023-02-08
Fix flaky test of test_rejectmap_mul_db

fanfuxiaoran:
There are some bugs in active tables, but we are not intended to fix them now. So just bring back the pause in the sql to make the test can run success at most time.

### DIFF
--- a/tests/regress/expected/test_rejectmap_mul_db.out
+++ b/tests/regress/expected/test_rejectmap_mul_db.out
@@ -22,8 +22,8 @@ SELECT diskquota.wait_for_worker_new_epoch();
 
 -- Trigger hard limit to dispatch rejectmap for tjmu1
 INSERT INTO b SELECT generate_series(1, 100000000); -- fail
-ERROR:  schema's disk space quota exceeded with name: 2200  (seg1 127.0.0.1:6003 pid=3985762)
--- NOTE: Pause to avoid tjmu1's worker clear the active table. Since the naptime is 0 on CI, this might be flaky.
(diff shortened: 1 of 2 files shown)


fanfuxiaoran:
There are some bugs in active tables, but we are not intended to fix them now. So just bring back the pause in the sql to make the test can run success at most time.

tests/regress/expected/test_rejectmap_mul_db.out
```diff
@@ -22,7 +22,14 @@ SELECT diskquota.wait_for_worker_new_epoch();
 
 -- Trigger hard limit to dispatch rejectmap for tjmu1
 INSERT INTO b SELECT generate_series(1, 100000000); -- fail
-ERROR:  schema's disk space quota exceeded with name: 2200  (seg1 127.0.0.1:8003 pid=43782)
+ERROR:  schema's disk space quota exceeded with name: 2200  (seg1 127.0.0.1:6003 pid=3985762)
+-- NOTE: Pause to avoid tjmu1's worker clear the active table. Since the naptime is 0 on CI, this might be flaky.
```
beeender, line 26: So, it will still be flaky, but a bit better than before, right? In that case, let's use `FIXME` instead of `NOTE`.

beeender (approved):

tests/regress/expected/test_rejectmap_mul_db.out
fanfuxiaoran: Ok

beeender (approved):

PR #303, opened 2023-02-09
Fix test_primary_failure test case

fanfuxiaoran:
add plpython_version variable in test sql.
On gpdb6, plpython_version is plpythonu, on gpdb7, it's plpython3u


zhrt123:
Should we use the name `test_primary_failure.in.sql`, which is the same as `test_rejectmap.in.sql`?

fanfuxiaoran:
> Should we use the name `test_primary_failure.in.sql`, which is the same as `test_rejectmap.in.sql`?

personally, I prefer this way. @beeender , how about you?

tests/regress/sql/test_primary_failure.sql
```diff
@@ -1,31 +1,43 @@
...
+\gset
+select CASE WHEN :gp_server_version_num < 70000 THEN 'plpythonu' else 'plpython3u' end plpython_version;
+\gset
+-- end_ignore
+
 CREATE SCHEMA ftsr;
 SELECT diskquota.set_schema_quota('ftsr', '1 MB');
 SET search_path TO ftsr;
-create or replace language plpythonu;
+create or replace language :plpython_version;
```
beeender, line 12: Let's use the way we introduced in #304  , so we can save lines 1-7.

Let me know if you don't have time to do that, I can help.

fanfuxiaoran: Done

beeender (approved):

beeender (changes requested):
Wait,
`diskquota_schedule7` should be deleted now. And the `EXPECTED_DIR_SUFFIX` in the `tests/CMakeLists.txt` can be removed as well.

fanfuxiaoran:
> Wait, `diskquota_schedule7` should be deleted now. And the `EXPECTED_DIR_SUFFIX` in the `tests/CMakeLists.txt` can be removed as well.

Ok, I'll fix it by the way.

beeender (approved):

PR #304, opened 2023-02-09
VAR replace for Regress.cmake and fix test_rejectmap

beeender:
Since GP7 doesn't support plpythonu, the test_rejectmap was broken for
GP7. This commit:

- Improve the Regress.cmake, so if the input sql file has a "in.sql"
  extension, "@VAR@" in it will be replaced by the corresponding cmake
  VAR.
- Set different @PLPYTHON_LANG_STR@ for GP6 & GP7
- Due to plpython composite type behaviour change, the python code in
  the test has been modified. The behaviour change is probably related
  to PG commit 94aceed317.

Also:
- Since most output files are the same for GP6 and GP7, just use
  symbolic links for them.


cmake/Regress.cmake
```diff
@@ -72,6 +75,21 @@ function(_Find_FaultInjector_Tests sql_DIR)
     set(fault_injector_test_list ${fault_injector_test_list} PARENT_SCOPE)
 endfunction()
 
+# Create symbolic links in the binary dir to input SQL files.
```
fanfuxiaoran, line 78: Good point! It's good to use the same sql or expect output file for gpdb7 and gpdb6. How about only creating output files that are different for gpdb6 and when running the tests,  we can copy the output files to the expected folder? 
So we can easily know which are different between gpdb6 and gpdb7.

cmake/Regress.cmake
```diff
@@ -25,6 +25,9 @@
 #  - regress_show_diff.sh
 #  - regress_loop.sh
 #
+# NOTE: If the input sql file extension is ".in.sql" instead of ".sql", the "@VAR@" in the input
```
fanfuxiaoran, line 28: Magic! I thought the source file is needed.

tests/isolation2/sql/test_rejectmap.in.sql
```diff
@@ -275,7 +284,7 @@ CREATE OR REPLACE FUNCTION replace_oid_with_relname(given_name text, filename te
       REGEXP_REPLACE(given_name,                                                          /*in func*/
          '^(pg_toast_|pg_aoseg_|pg_aovisimap_|pg_aoblkdir_|pg_aocsseg_)\d+',              /*in func*/
          '\1' ||                                                                          /*in func*/
-	 (SELECT DISTINCT relname FROM read_relation_cache_from_file(filename)            /*in func*/
+         (SELECT DISTINCT relname FROM read_relation_cache_from_file(filename)            /*in func*/
           WHERE  REGEXP_REPLACE(given_name, '\D', '', 'g') <> ''
```
higuoxing, line 288: Not an issue of this change, but it seems that `/*in func*/`  is missing in this line?

higuoxing:
There's another test case using `plpythonu`.

https://github.com/greenplum-db/diskquota/blob/b2367939bc2ff24bb8959ab6990e86ac559e7ef7/tests/regress/sql/test_primary_failure.sql#L4

cmake/Regress.cmake
beeender: Can do, see new commits

tests/isolation2/sql/test_rejectmap.in.sql
beeender: fixed

beeender:
> There's another test case using `plpythonu`.
> 
> [](https://github.com/greenplum-db/diskquota/blob/b2367939bc2ff24bb8959ab6990e86ac559e7ef7/tests/regress/sql/test_primary_failure.sql#L4)

This will be fixed in another PR.

higuoxing (approved):
Change looks good to me. I tried to rewrite plpython codes in plpgsql to avoid depending on plpython but failed due to some limitation of Greenplum and the plpgsql code is hard to read.

zhrt123 (approved):
LGTM. test_primary_failure also should be fixed.

PR #314, opened 2023-03-16
Fix bug: bgworker enters infinite loop after receiving notices from QE.

zhrt123:
### Problem
A dead loop in bgworker will happen following these steps:

- Bgworker executes a query by SPI_execute().
- Some QE's generate notices(for example, the crash of this server) and pass them to QD(bgworker).
- The bgworker wants to forward notices to the client in SPI_execute(). But there's no client for bgworker to establish the pq connection, which will cause a dead loop for bgworker.

The diskquota bgworker will always wait for internal_putbytes(), and can not update the table size.

### Analyze
The problem is caused by initializing MyProcPort in bgworker. In GPDB, if `MyProcPort == NULL`, the notice will never be forwarded to the client.
```
/*
* If MyProcPort is NULL, there is no client, so no need to generate notice.
* One example is that there is no client for a background worker.
*/
if (!res || MyProcPort == NULL) 
return;
```

### Solution
We should not initialize the MyProcPort in bgworker, which will cause the database name to not be automatically added to the log message. In normal, the log printed from the bgworker does not contain the database name but contains the bgworker's pid. In order to facilitate debugging, we should print the database name every BGWORKER_LOG_TIME to ensure that we can find the database name by the bgworker's pid in the log file.

### Test
- The relevant test case is built in `test_primary_failure`.
- `test_readiness_logged` is disabled because this test relies on `MyProcPort`.

### DIFF
--- a/tests/regress/diskquota_schedule
+++ b/tests/regress/diskquota_schedule
@@ -1,6 +1,7 @@
 test: config
 test: test_create_extension
-test: test_readiness_logged
+# This test seems to depend on MyProcPort, let's skip this test firstly.
+# test: test_readiness_logged
 test: test_init_table_size_table
 test: test_relation_size
 test: test_relation_cache
--- a/tests/regress/diskquota_schedule
+++ b/tests/regress/diskquota_schedule
@@ -1,7 +1,7 @@
 test: config
 test: test_create_extension
 # This test seems to depend on MyProcPort, let's skip this test firstly.
-# test: test_readiness_logged
+test: test_readiness_logged
 test: test_init_table_size_table
 test: test_relation_size
 test: test_relation_cache
--- a/tests/regress/diskquota_schedule
+++ b/tests/regress/diskquota_schedule
@@ -1,6 +1,5 @@
 test: config
 test: test_create_extension
-# This test seems to depend on MyProcPort, let's skip this test firstly.
 test: test_readiness_logged
 test: test_init_table_size_table
 test: test_relation_size
--- a/tests/regress/expected/test_readiness_logged.out
+++ b/tests/regress/expected/test_readiness_logged.out
@@ -1,5 +1,27 @@
 CREATE DATABASE test_readiness_logged;
 \c test_readiness_logged
+-- Get bgworker's log by database name.
+-- 1. select bgworker pid by database name.
+-- 2. select logmessage by bgworker pid.
+CREATE VIEW logmessage_count_view AS WITH logp AS(
+    SELECT
+        MAX(logpid) as max_logpid
+    FROM
+        gp_toolkit.gp_log_system
+    WHERE
+        position(
+            '[diskquota] start disk quota worker process to monitor database' in logmessage
+        ) > 0
+        AND position(current_database() in logmessage) > 0
+)
+SELECT
+    count(*)
+FROM
+    gp_toolkit.gp_log_system,
+    logp
+WHERE
+    logmessage = '[diskquota] diskquota is not ready'
+    and logpid = max_logpid;
 CREATE TABLE t (i int) DISTRIBUTED BY (i);
 CREATE EXTENSION diskquota;
 WARNING:  [diskquota] diskquota is not ready because current database is not empty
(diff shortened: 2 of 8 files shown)


zhrt123:
> A detailed PR message is needed, which should include:
> 
> 1. What was the problem, and the relevant gpdb source code
> 2. Why did we have these two line before
> 3. What is the side-effect by removing those two lines.

I'll add a detailed description.

beeender:
> Do we need a test case to make sure errors are reported correctly on bgworker?
> This can probably be done with gp_inject_fault().

I think this is not necessary. Reporting errors during SPI call is something should be ensured by GPDB, and if we need a test case, the case should exist in GPDB.

zhrt123:
> Do we need a test case to make sure errors are reported correctly on bgworker?
> 
> This can probably be done with gp_inject_fault().

My reply is lost.... The error report is logged in the log file and this procedure happens in GPDB, so it is difficult to make a test for this case.

zhrt123:
> > > Do we need a test case to make sure errors are reported correctly on bgworker?
> > > This can probably be done with gp_inject_fault().
> > 
> > 
> > I think this is not necessary. Reporting errors during SPI call is something should be ensured by GPDB, and if we need a test case, the case should exist in GPDB.
> 
> I mean to ensure that the bgworker of Diskquota functions properly when an error is reported from segments

We can add a report(WARNING) in a hook with fault_injector, and check whether the diskquota works normally after injecting a fault.

tests/regress/diskquota_schedule
```diff
@@ -1,6 +1,7 @@
 test: config
 test: test_create_extension
-test: test_readiness_logged
+# This test seems to depend on MyProcPort, let's skip this test firstly.
+# test: test_readiness_logged
```
beeender, line 4: Are you planning to solve this in the next PR?

If it is not too much effort to fix this test, it will be good to solve it in this PR.

zhrt123: I think it should be removed because it is not useful after fixing this bug....

tests/regress/expected/test_readiness_logged.out
```diff
@@ -1,5 +1,27 @@
...
+CREATE VIEW logmessage_count_view AS WITH logp AS(
+    SELECT
+        MAX(logpid) as max_logpid
+    FROM
+        gp_toolkit.__gp_log_master_ext
+    WHERE
+        position(
+            '[diskquota] start disk quota worker process to monitor database' in logmessage
+        ) > 0
+        AND position(current_database() in logmessage) > 0
```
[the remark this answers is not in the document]
zhrt123: I want to select the most recent log message, but after my experiment, it seems to be unable to achieve that.
```
zhrt=# SELECT                     
        logpid,logmessage
  FROM
        gp_toolkit.__gp_log_master_ext
    WHERE
        position(
            '[diskquota] start disk quota worker process to monitor database' in logmessage
        ) > 0
        AND position(current_database() in logmessage) > 0;
 logpid |                              logmessage                              
--------+----------------------------------------------------------------------
 p12916 | [diskquota] start disk quota worker process to monitor database:zhrt
 p14553 | [diskquota] start disk quota worker process to monitor database:zhrt
 p16843 | [diskquota] start disk quota worker process to monitor database:zhrt
 p44801 | [diskquota] start disk quota worker process to monitor database:zhrt
 p44801 | [diskquota] start disk quota worker process to monitor database:zhrt
(5 rows)

zhrt=# SELECT
        logpid,logmessage
  FROM
        gp_toolkit.__gp_log_master_ext
    WHERE
        position(
            '[diskquota] start disk quota worker process to monitor database' in logmessage
        ) > 0
        AND position(current_database() in logmessage) > 0 limit 1;
 logpid |                              logmessage                              
--------+----------------------------------------------------------------------
 p12916 | [diskquota] start disk quota worker process to monitor database:zhrt
(1 row)
```

tests/regress/diskquota_schedule
```diff
@@ -1,5 +1,6 @@
 test: config
 test: test_create_extension
+# This test seems to depend on MyProcPort, let's skip this test firstly.
```
beeender, line 3: remove this comment then it is good to og

PR #316, opened 2023-03-17
Fix regress result for GPDB7

zhrt123:
Some notices are removed from GPDB7, so the diskquota test result should be updated.


zhrt123:
This issue is fixed by #315 , close this PR.

PR #321, opened 2023-03-29
Skip refresh_disk_quota_model() when receiving a signal

zhrt123:
## Problem

When with often signals receiving diskquota may call refresh_disk_quota_model frequently than diskquota_naptime.

## Solution
When receiving a signal, the time waiting for latch `sleep_time` will be smaller than `diskquota_naptime`. We should skip refresh_disk_quota_model() when `sleep_time < diskquota_naptime`.

### DIFF
--- a/src/diskquota.c
+++ b/src/diskquota.c
@@ -540,9 +540,9 @@ disk_quota_worker_main(Datum main_arg)
 			MemoryAccounting_Reset();
 #endif /* GP_VERSION_NUM */
 
-			loop_start_timestamp = GetCurrentTimestamp();
-			sleep_time           = diskquota_naptime * 1000;
+			sleep_time = diskquota_naptime * 1000;
 		}
+		loop_start_timestamp = GetCurrentTimestamp();
 
 		if (DiskquotaLauncherShmem->isDynamicWorker)
 		{
--- a/src/diskquota.c
+++ b/src/diskquota.c
@@ -510,7 +510,10 @@ disk_quota_worker_main(Datum main_arg)
 		}
 
 		/*
-		 * diskquota model is refreshed every naptime. If the latch is set ahead of time,
+		 * If the bgworker receives a signal, the latch will be set ahead of the diskquota.naptime.
+		 * To avoid too frequent diskquota refresh caused by receiving the signal, we use
+		 * loop_start_timestamp and loop_end_timestamp to maintain the elapsed time since the last
+		 * diskquota refresh. If the latch is set ahead of diskquota.naptime,
 		 * refresh_disk_quota_model() should be skipped.
 		 */
 		loop_end_timestamp = GetCurrentTimestamp();
--- a/src/diskquota.c
+++ b/src/diskquota.c
@@ -485,6 +485,11 @@ disk_quota_worker_main(Datum main_arg)
 	bool        is_gang_destroyed   = false;
 	TimestampTz log_start_timestamp = GetCurrentTimestamp();
 	TimestampTz log_end_timestamp;
+	TimestampTz loop_start_timestamp = 0;
+	TimestampTz loop_end_timestamp;
+	long        sleep_time = 0;
+	long        secs;
+	int         usecs;
 	ereport(LOG, (errmsg("[diskquota] disk quota worker process is monitoring database:%s", dbname)));
 
 	while (!got_sigterm)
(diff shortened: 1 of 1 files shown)


src/diskquota.c
```diff
@@ -534,8 +553,7 @@ disk_quota_worker_main(Datum main_arg)
 		 * necessary, but is awakened if postmaster dies.  That way the
 		 * background process goes away immediately in an emergency.
 		 */
-		rc = DiskquotaWaitLatch(&MyProc->procLatch, WL_LATCH_SET | WL_TIMEOUT | WL_POSTMASTER_DEATH,
```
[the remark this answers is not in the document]
zhrt123: YES, we can check `rc`, but we also need to calculate the next sleep_time.

src/diskquota.c
```diff
@@ -501,27 +506,41 @@ disk_quota_worker_main(Datum main_arg)
...
-			is_gang_destroyed             = false;
-		}
-		else if (!is_gang_destroyed)
+		/*
+		 * diskquota model is refreshed every naptime. If the latch is set ahead of time,
+		 * refresh_disk_quota_model() should be skipped.
+		 */
+		loop_end_timestamp = GetCurrentTimestamp();
+		TimestampDifference(loop_start_timestamp, loop_end_timestamp, &secs, &usecs);
+		sleep_time -= secs * 1000 + usecs / 1000;
```
fanfuxiaoran, line 515: As the sleep_time is updated, the loop_start_time should be updated at the same time. Otherwise, the sleep_time will be not correct.

fanfuxiaoran:
I'm still not sure if the signal will trigger the refresh action frequently and if it's worth complexing the code here to handle it.

zhrt123:
> I'm still not sure if the signal will trigger the refresh action frequently and if it's worth complexing the code here to handle it.

In the multi-user scene,  it may cause frequent refresh action.

zhrt123:
> Could we have a test case for this?

I have no idea now.

src/diskquota.c
```diff
@@ -504,27 +509,41 @@ disk_quota_worker_main(Datum main_arg)
...
-		SIMPLE_FAULT_INJECTOR("diskquota_worker_main");
-		if (!diskquota_is_paused())
-		{
-			/* Refresh quota model with init mode */
-			refresh_disk_quota_model(!MyWorkerInfo->dbEntry->inited);
-			MyWorkerInfo->dbEntry->inited = true;
-			is_gang_destroyed             = false;
-		}
-		else if (!is_gang_destroyed)
+		/*
```
fanfuxiaoran, line 512: Could you add more comments around the loop time or signal-related codes?  It's really easy to get lost in those codes and cannot catch the main idea of it. Or is it possible to separate the timeout computation code from the main function body?

zhrt123: I will add more comments on these codes.

src/diskquota.c
```diff
@@ -504,27 +509,44 @@ disk_quota_worker_main(Datum main_arg)
...
```
fanfuxiaoran, line 512: > log_end_timestamp = GetCurrentTimestamp();

Although this is not added in the current pr,  a little strange.  Can it lead to too many logs, if it does, we can just log it once.

src/diskquota.c
```diff
@@ -563,8 +584,6 @@ disk_quota_worker_main(Datum main_arg)
 		ereport(LOG, (errmsg("[diskquota] stop disk quota worker process to monitor database:%s", dbname)));
 	ereport(DEBUG1, (errmsg("[diskquota] stop disk quota worker process to monitor database:%s", dbname)));
```
fanfuxiaoran, line 585: Is the ereport correct here?

fanfuxiaoran:
LGTM!

zhrt123:
> > > Could we have a test case for this?
> > 
> > 
> > I have no idea now.
> 
> @zhrt123 Any update on this?

Not yet. I think it is not necessary to test this time duration case.

src/diskquota.c
zhrt123: No, we can not remove it. We should log the database name every hour to find which database throws the error by database oid. If we only log it once, the log file that records the database name may be deleted after a long time.

src/diskquota.c
zhrt123: I tink it is not problem.

src/diskquota.c
```diff
@@ -518,8 +518,8 @@ disk_quota_worker_main(Datum main_arg)
 		 */
 		loop_end_timestamp = GetCurrentTimestamp();
 		TimestampDifference(loop_start_timestamp, loop_end_timestamp, &secs, &usecs);
-		sleep_time -= secs * 1000 + usecs / 1000;
-		if (sleep_time <= 0)
+		sleep_time += secs * 1000 + usecs / 1000;
```
fanfuxiaoran, line 521: any overflows here?

src/diskquota.c
```diff
@@ -518,8 +518,8 @@ disk_quota_worker_main(Datum main_arg)
...
+		if (sleep_time >= diskquota_naptime * 1000)
```
fanfuxiaoran, line 522: It looks conflicted with `long sleep_time = diskquota_naptime;`

src/diskquota.c
zhrt123: No, if the sleep_time overflows, time will pass billions of years.

src/diskquota.c
zhrt123: `long sleep_time = diskquota_naptime;` is used to guarantee the first refresh will be executed without sleep.

fanfuxiaoran: Yeah, I know the meaning of it, but it's better to set it to diskquota_naptime * 1000 as that's a little confusing that whether it is seconds or micro seconds.

Issue #318, opened 2023-03-23
string may too big


RekGRpth:
how about instead of constructing large string in https://github.com/greenplum-db/diskquota/blob/a939bc078f9b041059ecb223504abb816753a3e4/src/quotamodel.c#L1158-L1190 use view and it implementation like pg_stat_activity and pg_stat_get_activity to avoid such errors as
```sql
Cannot enlarge string buffer containing 1073741807 bytes by 20 more bytes.
```
?

RekGRpth:
> maybe we should use `COPY`

how can we use `copy` for deleting?

RekGRpth:
> in your use case, it is the `deleted_table_expr`

yes

RekGRpth:
> why using views can mitigate this issue?

because with view one may execute something like
```sql
delete from diskquota.table_size where (tableid, segid) in ( SELECT * FROM deleted_table_view );
```
where deleted_table_view is view, realized via c-function like pg_stat_activity and pg_stat_get_activity

RekGRpth:
at start diskquota tries to load all table into hash map in ram https://github.com/greenplum-db/diskquota/blob/a939bc078f9b041059ecb223504abb816753a3e4/src/gp_activetable.c#L939
but in out case it fails with error
```csv
"ERROR","XX000","invalid memory alloc request size 1073741824 (context 'SPI TupTable') (mcxt.c:1357)",,,,,"SQL statement ""select tableid, size, segid from diskquota.table_size""",,0,,"mcxt.c",571,
```
but this error is catched by https://github.com/greenplum-db/diskquota/blob/a939bc078f9b041059ecb223504abb816753a3e4/src/quotamodel.c#L796-L845 thus diskquota simply ignores ignores unsuccessful loading and filling hash map and thinks that it contains nothing and forms sql to remove all ids from table, but again exceeds max alloc size
```sql
Cannot enlarge string buffer containing 1073741807 bytes by 20 more bytes.
```

RekGRpth:
instead load all table to ram at once I suggest doing something like in pg_prewarm extension

zhrt123:
```
"ERROR","XX000","invalid memory alloc request size 1073741824 (context 'SPI TupTable') (mcxt.c:1357)",,,,,"SQL statement ""select tableid, size, segid from diskquota.table_size""",,0,,"mcxt.c",571,
```
[...]

PR #322, opened 2023-03-30
Replace SPI_execute with SPI_cursor_fetch

zhrt123:
## Problem
Issue: #318 
The result buffer of SPI_execute is limited to 1GB. If the number of rows in `diskquota.table_size` exceeds 3 * 10^7 (500 segments with 60000 tables for example), `SPI_execute("select tableid, size, segid from diskquota.table_size")` will raise an error for `invalid memory alloc request size 1073741824 (context 'SPI TupTable')`.

## Solution

We should fetch table-size entries by portions. Replace SPI_execute() with SPI_cursor_fetch() in load_table_size().

### DIFF
--- a/src/gp_activetable.c
+++ b/src/gp_activetable.c
@@ -936,16 +936,26 @@ get_active_tables_oid(void)
 static void
 load_table_size(HTAB *local_table_stats_map)
 {
-	int                        ret;
 	TupleDesc                  tupdesc;
 	int                        i;
 	bool                       found;
 	TableEntryKey              key;
 	DiskQuotaActiveTableEntry *quota_entry;
+	SPIPlanPtr                 plan;
+	Portal                     portal;
+	char                      *sql = "select tableid, size, segid from diskquota.table_size";
 
-	ret = SPI_execute("select tableid, size, segid from diskquota.table_size", true, 0);
-	if (ret != SPI_OK_SELECT)
-		ereport(ERROR, (errmsg("[diskquota] load_table_size SPI_execute failed: return code %d, error: %m", ret)));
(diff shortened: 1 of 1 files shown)


src/gp_activetable.c
```diff
@@ -936,16 +936,26 @@ get_active_tables_oid(void)
...
-		ereport(ERROR, (errmsg("[diskquota] load_table_size SPI_execute failed: return code %d, error: %m", ret)));
+	if ((plan = SPI_prepare(sql, 0, NULL)) == NULL)
+		ereport(ERROR, (errmsg("[diskquota] SPI_prepare(\"%s\") failed", sql)));
+	if ((portal = SPI_cursor_open(NULL, plan, NULL, NULL, true)) == NULL)
+		ereport(ERROR, (errmsg("[diskquota] SPI_cursor_open(\"%s\") failed", sql)));
+
+	SPI_cursor_fetch(portal, true, 10000);
+
+	if (SPI_tuptable == NULL)
+	{
```
higuoxing, line 956: Do we need to call `SPI_cursor_close()` before error out?

higuoxing (approved):
LGTM with one question.

src/gp_activetable.c
zhrt123: Reference to GPDB's code, we do not need to call this function.

fanfuxiaoran (approved):

fanfuxiaoran:
Better to add some corner test cases for it: such as there are 0 or 10000 , 20000 rows in the table_size

zhrt123:
> Better to add some corner test cases for it: such as there are 0 or 10000 , 20000 rows in the table_size

Since this test will be ver slow, we'd better to test this case at local.

Issue #323, opened 2023-03-30
You should not pass database name via background worker structure


RekGRpth:
You should not pass database name via background worker structure
https://github.com/greenplum-db/diskquota/blob/a939bc078f9b041059ecb223504abb816753a3e4/src/diskquota.c#L323
because it is ascii-filtered in https://github.com/greenplum-db/gpdb/blob/f182228992b62e2023e2fac5b4971406abd35c9d/src/backend/postmaster/bgworker.c#L385-L386 or in https://github.com/greenplum-db/gpdb/blob/29d2bbd37779ad54cdf922ba8dfaf42cb1678725/src/backend/postmaster/bgworker.c#L300-L301
thus, if database name contains not-ascii symbols then diskquota does not work with it

PR #326, opened 2023-04-03
Add dbname to DiskQuotaWorkerEntry

zhrt123:
## Problem
Previously, we passed dbname by bgw_name to bgworker. But refer to https://github.com/greenplum-db/gpdb/blob/f182228992b62e2023e2fac5b4971406abd35c9d/src/backend/postmaster/bgworker.c#L385-L386, bgw_name is copied by ascii_safe_strlcpy(), which replaces non-ASCII bytes with '?'. If the database name is non-ASCII, the bgworker can not connect to the correct database and raises an error. #323 

## Solution
Pass dbname by bgw_main_arg.


src/diskquota.c
```diff
@@ -1381,7 +1384,11 @@ start_worker(DiskquotaDBEntry *dbEntry)
 		result = INVALID_DB;
 		goto Failed;
 	}
-	snprintf(worker.bgw_name, sizeof(worker.bgw_name), "%s", dbname);
+	LWLockAcquire(diskquota_locks.workerlist_lock, LW_EXCLUSIVE);
+	namestrcpy(&(dq_worker->dbname), dbname);
```
fanfuxiaoran, line 1388: LGTM.  However,  the dbname is not needed to start the bgworker, so we should not get it here and pass it to diskquota worker through shared memory. Instead, the diskquota worker can get it through `get_db_name` by itself.

src/diskquota.c
```diff
@@ -1381,7 +1384,11 @@ start_worker(DiskquotaDBEntry *dbEntry)
 		result = INVALID_DB;
 		goto Failed;
 	}
-	snprintf(worker.bgw_name, sizeof(worker.bgw_name), "%s", dbname);
+	LWLockAcquire(diskquota_locks.workerlist_lock, LW_EXCLUSIVE);
```
beeender, line 1387: Why lock is needed here but not at line 1361?

src/diskquota.c
beeender: This sounds better, can we do that?

beeender:
BTW, since this is related to #323 , it is better we have the issue number in the PR/commit message.

src/diskquota.c
zhrt123: Can the bgworker run get_db_name() before BackgroundWorkerInitializeConnection()?

src/diskquota.c
zhrt123: Maybe we can remove it.

zhrt123:
> This sounds better, can we do that?

We should give bgworker a name.

src/diskquota.c
fanfuxiaoran: >Can the bgworker run get_db_name() before BackgroundWorkerInitializeConnection()?

fanfuxiaoran: In gpdb7,  a filed ` char            bgw_extra[BGW_EXTRALEN];` is added, we can use it to pass dbname.

src/diskquota.c
```diff
@@ -321,7 +321,12 @@ define_guc_variables(void)
 void
 disk_quota_worker_main(Datum main_arg)
 {
-	char *dbname = MyBgworkerEntry->bgw_name;
+	char dbname[NAMEDATALEN];
+
+	MyWorkerInfo = (DiskQuotaWorkerEntry *)DatumGetPointer(MyBgworkerEntry->bgw_main_arg);
+	Assert(MyWorkerInfo != NULL);
+
+	memcpy(dbname, MyWorkerInfo->dbname.data, NAMEDATALEN);
```
fanfuxiaoran, line 329: Can use the `MyWorkerInfo->dbname.data` directly instead of copying it? And can use `NameStr` to get the dbname.data.

fanfuxiaoran (approved):

src/diskquota.c
fanfuxiaoran: after discussion, we decide to save the dbname in shared memory as the pr did to keep it has the same behaviors between gpdb6 and gpdb7.

fanfuxiaoran (approved):

tests/regress/expected/test_bgworker_name.out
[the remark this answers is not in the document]
zhrt123: Renamed

src/diskquota.c
beeender: one change

```
const char* dbname = MyWorkerInfo->dbname.data;
```

zhrt123: I think it is better to use local memory to store dbname.

yihong0618 (approved):

beeender (approved):

zhrt123:
Previously, we passed dbname by bgw_name to bgworker. But refer to https://github.com/greenplum-db/gpdb/blob/f182228992b62e2023e2fac5b4971406abd35c9d/src/backend/postmaster/bgworker.c#L385-L386, bgw_name is copied by ascii_safe_strlcpy(), which replaces non-ASCII bytes with '?'. If the database name is non-ASCII, the bgworker can not connect to the correct database and raises an error. issue: https://github.com/greenplum-db/diskquota/issues/323 . To solve this problem, we should pass dbname by bgw_main_arg to the main function of the bgworker.

PR #338, opened 2023-05-09
Upgrade to 3.0

zhrt123:


zhrt123:
> 

It is a good idea. We can discuss how to achieve it.

zhrt123:
> Can we make `control/test/diskquota_test--1.0.sql`, `control/ddl/diskquota.control` and `control/ddl/diskquota--XX.sql` into templates?
> 
> So that we don't need to modify the version string in `control/test/diskquota_test--1.0.sql`, `control/ddl/diskquota.control` and we don't need to copy the whole file of `control/ddl/diskquota--XX.sql` each time we bumping a new version.


It is a good idea. We can discuss how to achieve it.

zhrt123:
The failed test for GPDB is normal, we can ignore it in this PR.

PR #341, opened 2023-06-13
fix creating extension with global option appendonly=true

red1452:
All tables in diskquota use primary keys. Unique index tables are created for primary keys. Unique index tables can not be created for table with option appendonly=true. appendonly=false has been added for each table with primary key.

## Steps to reproduce.
1) build diskquota extension
2) append it to shared_preload_libraries
3) create diskquota database
4) restart cluster
5) execute in the postgres (on any other) database:
```
set gp_default_storage_options='appendonly=true';
create extension diskquota;
```
## Result
```
ERROR:  append-only tables do not support unique indexes
```

### DIFF
--- a/control/ddl/diskquota--1.0--2.0.sql
+++ b/control/ddl/diskquota--1.0--2.0.sql
@@ -9,7 +9,7 @@ CREATE TABLE diskquota.target (
 	primaryOid oid,
 	tablespaceOid oid, -- REFERENCES pg_tablespace.oid,
 	PRIMARY KEY (primaryOid, tablespaceOid, quotatype)
-);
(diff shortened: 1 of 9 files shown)


zhrt123:
LGTM, I will do some tests on CI and then merge it. Thanks very much.

zhrt123:
We won't update the extension sql file of the previous version. We will add the modification on diskquota--2.3.sql.

zhrt123:
We created a new PR for this fix. Thanks @red1452

zhrt123:
Let's close this PR and see the discussion in #366

zhrt123:
Let's close this PR and see the discussion in #366

PR #343, opened 2023-06-21
Fix use-after-free introduced in PR339.

higuoxing:
#339 introduced a use-after-free bug to diskquota. The root cause is that the `TextDatumGetCString()` returns a pallocated string and it got freed when the memory context is gone.


higuoxing:
In favor of #344. Close.

PR #357, opened 2023-07-04
Update resources by using gp-extensions-ci subtree

liuxueyang:
Migrage changes https://github.com/pivotal/pgaudit/pull/13 here


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "concourse/lib"]
+	path = concourse/lib
+	url = ../gp-extensions-ci.git
```
beeender, line 3: this won't work since diskquota has a different org

`greenplum-db/diskquota` vs `pivotal/gp-extensions-ci`

higuoxing:
I'm curious, is there any specific reason that we make the `gp-extensions-ci` repo as a submodule or a subtree of extensions? Can we just leave it as a standalone repo?

Currently, each time we modify the CI scripts, we need to do the following things:

- Modify the `gp-extensions-ci` repo.
- Bump a new version for the `gp-extensions-ci submodule` in other extensions' repo.
- Set the concourse pipeline.

I would like to make `gp-extensions-ci` as a standalone repo. Then, each time we modify the CI scripts, we can save some steps.

- Modify the gp-extensions-ci repo.
- Set the concourse pipeline (Maybe we can have some handy scripts to do it for us. E.g.,)
  - Set the diskquota PR pipeline, we just need to type the following commands in the `gp-extensions-ci` repo: `make set-diskquota-pr`.

beeender (approved):

higuoxing (approved):

PR #359, opened 2023-07-14
Error handling for disqkuota worker startup stage.

bimboterminator1:
During diskquota worker's first run the initial set of active tables with their sizes is being loaded from `diskquota.table_size` table in order to warm up diskquota rejectmap and other shared memory objects. If an error occurs during this initialization process, the error will be ignored in PG_CATCH() block. Such ignorance can be potentially harmful and can lead to undesired behaviour of the whole extension. For example, if an error ocurs during initialization, `local_active_table_stat_map` will not be filled properly. And at the next loop iteration, tables, that are not in acitive table list will be marked as irrelevant and to be deleted both from `table_size_map` and `table_size` table in `flush_to_table_size` function. This situation produces extra perfomance load, which is not guaranteed to be safe.

This commit proposes the handling of the initialization errors, which occur during worker's first run. In the `DiskquotaDBEntry` structure the bool variable `corrupted` is added in order to indicate, that the worker wasn't able to initialize itself on given database. And `DiskquotaDBEntry` also is now passed to `refresh_disk_quota_model` function from worker main loop, because one need to change the state of dbEntry. The state is changed when the `refresh_disk_quota_usage` function catches an error, which occured during the initialization step, in PG_CATCH() block. And after the error is catched, the `corrupted` flag is set in given dbEntry, and then the error is rethrown. This leads to worker process termination. The launcher will not be able to start it again, because added flag is set in the database structure, and this flag is being checked inside the `disk_quota_launcher_main` function. The flag can be reseted by calling `resetBackgroundWorkerCorruption` function, which is currently called in SIGHUP handler.

This patch does not contain a proper test because the author couldn't find optimal architecture for the test. The behaviour of the patch can be testes either via putting fault injector or calling an error, for example, in the  [gp_fetch_active_tables](https://github.com/greenplum-db/diskquota/blob/f9e940fbaf3dbec7f156b5c7f2c3b3a384c4dd0b/src/gp_activetable.c#L381) function. The presence of diskquota bgworker process can be monitored via `ps` or `pg_stat_activity` for given database.

### DIFF
--- a/src/diskquota.c
+++ b/src/diskquota.c
@@ -144,6 +144,7 @@ static void                    vacuum_db_entry(DiskquotaDBEntry *db);
 static void                    init_bgworker_handles(void);
 static BackgroundWorkerHandle *get_bgworker_handle(uint32 worker_id);
 static void                    free_bgworker_handle(uint32 worker_id);
+static void                    resetBackgroundWorkerCorruption(void);
 #if GP_VERSION_NUM < 70000
 /* WaitForBackgroundWorkerShutdown is copied from gpdb7 */
 static BgwHandleStatus WaitForBackgroundWorkerShutdown(BackgroundWorkerHandle *handle);
@@ -575,7 +576,7 @@ disk_quota_worker_main(Datum main_arg)
 			if (!diskquota_is_paused())
 			{
 				/* Refresh quota model with init mode */
-				refresh_disk_quota_model(!MyWorkerInfo->dbEntry->inited);
+				refresh_disk_quota_model(MyWorkerInfo->dbEntry);
 				MyWorkerInfo->dbEntry->inited = true;
 				is_gang_destroyed             = false;
 			}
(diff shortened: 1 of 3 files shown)


zhrt123:
Thanks very much. This is an excellent enhancement for diskquota. But IMHO, resetting `corrupted` when receiving SIGHUP needs to be further considered since this signal could be sent after a GUC value is set. Can we just rethrow the error, print the error log, and wait for the next initialization of this db?

bimboterminator1:
Hi! Thanks for your reply. 
>  But IMHO, resetting `corrupted` when receiving SIGHUP needs to be further considered since this signal could be sent after a GUC value is set. 

Yes, I share your point of view. We don't need to try to relaunch the bgworker for corrupted db any time when the Greenplum reloads its configuration.
Therefore, there is a question, how we can reinitialize the worker at current db and affect the `corrupted` flag from outside? 

Or we can remove logic with `corrupted` flag and we may completely remove DiskquotaDBEntry from database list (both from DiskquotaLauncherShmem->dbArray and diskquota_namespace.database_list, but this will be too implicit for the user, it's the same thing as drop extension i guess)?

> We need to print more error messages to users so that they can handle this error. Can we rethrow the error, print the error log, and wait for the next initialization of this db?

Yep, I suggest print the original error first, then print next error message like "Diskquota worker initialization failed on db: dbname":

PR #362, opened 2023-08-03
Update to 2.3.0

zhrt123:
Fix the following issues:

diskquota cannot correctly print warning when calling `create extension` in a non-empty database. The diskquota.state shows whether the current database is empty. Previously, we update the status to diskquota.state during `create extension` and query the status in UDF `diskquota.diskquota_start_worker`. When querying the database status, the relations whose `relkind in ('v', 'c', 'f')` are skipped, while all relations are filtered when updating the database status. This patch merges the two SQL statements to `INSERT RETURNING` to solve this problem.
Benefit: we won't need to upgrade the minor version of diskquota when changing the statement.

Remove useless test `test_update_db_cache`. The behavior of the bgworker scheduler is changed, so this test is no longer useful.

Update the upgrade test script `2.2_set_quota`. There should be `select diskquota.init_table_size_table()` in `2.2_set_quota` since the above issue.

Add upgrade test for gp7. 

Fix upgrade test for gp6.


upgrade_test/CMakeLists.txt
```diff
@@ -14,15 +14,22 @@ if (${GP_MAJOR_VERSION} EQUAL 6)
 		${CMAKE_CURRENT_SOURCE_DIR}/schedule_2.1--2.2
 		${CMAKE_CURRENT_SOURCE_DIR}/schedule_2.2--2.1
 		)
+  set(EXPECTED_DIR "${CMAKE_CURRENT_SOURCE_DIR}/expected")
```
beeender, line 17: this is not right, I don't think this will pass the CI.

See the example in the `tests/CMakeLists.txt`, the `expected7` will be appended to the list to overwrite the same tests.

beeender: This is correct. 7 is from `2.2`

beeender (approved):

beeender:
I did a careful review, just merge this.

PR #366, opened 2023-08-24
fix creating extension with global option appendonly=true.

zhrt123:
# Problem
All tables in diskquota use primary keys. Unique index tables are created for primary keys. Unique index tables can not be created for tables with the option appendonly=true. This issue only occurs in gp6 since `set gp_default_storage_options='appendonly=true';` is disabled in gp7.

# Steps to reproduce.
1) build diskquota extension
2) append it to shared_preload_libraries
3) create diskquota database
4) restart cluster
5) execute in the postgres (on any other) database:
```
set gp_default_storage_options='appendonly=true';
create extension diskquota;
```
# Result
```
ERROR:  append-only tables do not support unique indexes
```

# Solution
Add appendonly=false to each table with the primary key.

This PR is refactored from #341 

Co-authored-by: Evgeniy Ratkov <e.ratkov@arenadata.io>

### DIFF
--- a/control/ddl/diskquota--2.3.sql
+++ b/control/ddl/diskquota--2.3.sql
@@ -11,7 +11,7 @@ CREATE TABLE diskquota.quota_config(
 	quotalimitMB int8,
 	segratio float4 DEFAULT 0,
 	PRIMARY KEY(targetOid, quotatype)
-) DISTRIBUTED BY (targetOid, quotatype);
+) WITH (appendonly=false) DISTRIBUTED BY (targetOid, quotatype);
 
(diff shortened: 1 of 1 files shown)


higuoxing (approved):
LGTM! BTW, The `Co-authored-by` section in commit message should be

```
Co-authored-by: Evgeniy Ratkov <e.ratkov@arenadata.io>
```

otherwise, github cannot recognize it.

beeender (approved):

red1452:
Please, change `Co-authored by` to `Co-authored-by`, to get correct "co-authored" section in commit message.

Issue #360, opened 2023-07-27
user may load diskquota library without shared memory initialization


Stolb27:
There are several commits (05da9d4 and 86ff586) in diskquota 2.2.2 that eased restrictions for extension initialization check to make it possible to update extension. But! These changes make it possible to run diskquota queries against uninitialized extension too. That leads to segmentation fault on segments as far as shared structures isn't initialized properly.

For example, let's create a stand with diskquota 2.1 and 2.2 shared libraries

## Steps to reproduce 1:

```
\! gpconfig -c shared_preload_libraries -v diskquota-2.1
\! gpstop -arfq
\c
create extension diskquota version '2.1';
\! gpconfig -c shared_preload_libraries -v ''
\! gpstop -arfq
\c
alter extension diskquota update;
drop extension diskquota;
create extension diskquota version '2.2';
```

## Step to reproduce 2:

```
\! gpconfig -c shared_preload_libraries -v diskquota-2.1
\! gpstop -arfq
\c
create extension diskquota version '2.1';
\! gpconfig -c shared_preload_libraries -v ''
\! gpstop -arfq
\c
alter extension diskquota update;
select diskquota.resume();
```

Expected behavior:

```
ERROR:  [diskquota] booting 2.2.2, but diskquota-2.2 not in shared_preload_libraries. abort. (diskquota.c:224)  (seg0 slice1 127.0.1.1:6002 pid=17043) (diskquota.c:224)
CONTEXT:  SQL statement "SELECT diskquota.diskquota_fetch_table_stat(5, '{794247}'::oid[]) FROM gp_dist_random('gp_id')"
```

zhrt123:
Thanks for your issue. We created a PR for this problem.

PR #368, opened 2023-08-30
Fix continuous upgrade

zhrt123:
# Problem
Several commits (05da9d4 and 86ff586) in diskquota 2.2.2 eased restrictions for extension initialization checks to make it possible to update extension. But these changes make it possible to run diskquota queries against uninitialized extension too. That leads to segmentation fault on segments as far as shared structures isn't initialized properly.

issue: #360 

# Solution
During the upgrade procedure, when diskquota.so is not in shared_preload_libraries, check whether the current version is the final version:
- If yes, throw an error to the user.
- Otherwise, return from _PG_init().

### DIFF
--- a/src/diskquota.c
+++ b/src/diskquota.c
@@ -148,7 +148,8 @@ static void                    free_bgworker_handle(uint32 worker_id);
 /* WaitForBackgroundWorkerShutdown is copied from gpdb7 */
 static BgwHandleStatus WaitForBackgroundWorkerShutdown(BackgroundWorkerHandle *handle);
 #endif /* GP_VERSION_NUM */
-static bool is_altering_extension(void);
+static bool is_altering_extension_to_default_version(char *version);
+static bool check_alter_extension(void);
 
 /*
  * diskquota_launcher_shmem_size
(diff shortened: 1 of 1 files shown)


src/diskquota.c
```diff
@@ -178,25 +208,43 @@ is_altering_extension(void)
...
-	char *pos;
-	bool  match = true;
+	char        *query  = asc_tolower(ActivePortal->sourceText, strlen(ActivePortal->sourceText));
+	char        *pos    = query;
+	bool         ret    = true;
+	static char *regs[] = {"alter", "extension", "diskquota", "update"};
+	int          i;
+
+	/* Check whether the sql statement is alter extension. */
+	for (i = 0; i < 4; i++)
```
Sasasu, line 218:
```suggestion
	for (i = 0; i < sizeof(regs) / sizeof(char*); i++)
```

Sasasu (approved):

src/diskquota.c
zhrt123: hmm, the division will be done 4 times.

Sasasu: `sizeof()` is const expression, and `32/4` is also a const expression. the division will be done at compile time.

PR #369, opened 2023-08-31
Add GUC diskquota.max_database.

zhrt123:
This GUC shows the max number of database diskquota can monitor.

### DIFF
--- a/src/diskquota.c
+++ b/src/diskquota.c
@@ -76,6 +76,7 @@ int  diskquota_worker_timeout     = 60; /* default timeout is 60 seconds */
 bool diskquota_hardlimit          = false;
 int  diskquota_max_workers        = 10;
 int  diskquota_max_table_segments = 0;
+int  diskquota_max_database       = 0;
 
 DiskQuotaLocks       diskquota_locks;
 ExtensionDDLMessage *extension_ddl_message = NULL;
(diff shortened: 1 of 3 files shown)


beeender (commented):
Add some simple tests then we are good to go.

src/diskquota.c
```diff
@@ -76,6 +76,7 @@ int  diskquota_worker_timeout     = 60; /* default timeout is 60 seconds */
 bool diskquota_hardlimit          = false;
 int  diskquota_max_workers        = 10;
 int  diskquota_max_table_segments = 0;
+int  diskquota_max_database       = 0;
```
fanfuxiaoran, line 79: How about init with the default value :1024 ?

src/diskquota.c
```diff
@@ -353,9 +355,10 @@ define_guc_variables(void)
 	        "Max number of backgroud workers to run diskquota extension, should be less than max_worker_processes.",
 	        NULL, &diskquota_max_workers, 10, 1, 20, PGC_POSTMASTER, 0, NULL, NULL, NULL);
 	DefineCustomIntVariable("diskquota.max_table_segments", "Max number of tables segments on the cluster.", NULL,
-	                        &diskquota_max_table_segments, 10 * 1024 * 1024,
-	                        INIT_NUM_TABLE_SIZE_ENTRIES * MAX_NUM_MONITORED_DB, INT_MAX, PGC_POSTMASTER, 0, NULL, NULL,
-	                        NULL);
+	                        &diskquota_max_table_segments, 10 * 1024 * 1024, INIT_NUM_TABLE_SIZE_ENTRIES * 1024,
```
fanfuxiaoran, line 358: Can we use `diskquota_max_database` instead of `1024` here in `INIT_NUM_TABLE_SIZE_ENTRIES * 1024`

zhrt123: We could do that. I'm not sure the sequence of the GUC definition would effect the result.

src/diskquota.c
zhrt123: The value is useless here.

Sasasu (approved):
the PR is fine to me.

the GUC effect sequence does exist.
the min value of `diskquota.max_table_segments` is the max value of `diskquota.max_database`.  I think this is fine.

src/diskquota.c
radarwave, line 79: Suggest using the name 'max_monitored_databases'.

fanfuxiaoran: Hmm, setting it to default value here can make the dikquota work even if it is not loaded with calling pg_init, but that's not a big deal.

fanfuxiaoran: Agree with 'max_monitored_databases'

fanfuxiaoran (approved):

src/diskquota.c
zhrt123: Changed.

PR #372, opened 2023-09-06
Add GUC diskquota.max_quotas

zhrt123:
`diskquota.max_quotas` shows the max number of quotas on the cluster.
In the current version of diskquota, quotas means:
- role
- schema
- role_tablespace
- schema_tablespace

The number of quota =
       role + schema + (role * tablespace) + (schema * tablespace).

This PR is waiting for #369

### DIFF
--- a/src/diskquota.c
+++ b/src/diskquota.c
@@ -77,6 +77,7 @@ bool diskquota_hardlimit               = false;
 int  diskquota_max_workers             = 10;
 int  diskquota_max_table_segments      = 0;
 int  diskquota_max_monitored_databases = 0;
+int  diskquota_max_quotas              = 0;
 
 DiskQuotaLocks       diskquota_locks;
 ExtensionDDLMessage *extension_ddl_message = NULL;
@@ -90,6 +91,9 @@ static int num_db = 0;
 
 /* how many TableSizeEntry are maintained in all the table_size_map in shared memory*/
 pg_atomic_uint32 *diskquota_table_size_entry_num;
+
+/* how many QuotaMapEntry are maintained in all the quota_info[type].map in shared memory*/
+pg_atomic_uint32 *diskquota_quota_info_entry_num;
 
 static DiskquotaLauncherShmemStruct *DiskquotaLauncherShmem;
 
@@ -407,6 +411,8 @@ define_guc_variables(void)
 	                        INT_MAX, PGC_POSTMASTER, 0, NULL, NULL, NULL);
 	DefineCustomIntVariable("diskquota.max_monitored_databases", "Max number of database on the cluster.", NULL,
 	                        &diskquota_max_monitored_databases, 50, 1, 1024, PGC_POSTMASTER, 0, NULL, NULL, NULL);
+	DefineCustomIntVariable("diskquota.max_quotas", "Max number of quotas on the cluster.", NULL, &diskquota_max_quotas,
+	                        1024 * 1024, 1024 * NUM_QUOTA_TYPES, INT_MAX, PGC_POSTMASTER, 0, NULL, NULL, NULL);
 }
 
 /* ---- Functions for disk quota worker process ---- */
@@ -1799,6 +1805,11 @@ init_launcher_shmem()
 	diskquota_table_size_entry_num =
 	        ShmemInitStruct("diskquota TableSizeEntry counter", sizeof(pg_atomic_uint32), &found);
 	if (!found) pg_atomic_init_u32(diskquota_table_size_entry_num, 0);
+
+	/* init QuotaInfoEntry counter */
+	diskquota_quota_info_entry_num =
+	        ShmemInitStruct("diskquota QuotaInfoEntry counter", sizeof(pg_atomic_uint32), &found);
+	if (!found) pg_atomic_init_u32(diskquota_quota_info_entry_num, 0);
 }
 
 /*
--- a/src/quotamodel.c
+++ b/src/quotamodel.c
@@ -55,8 +55,11 @@
 #define NUM_QUOTA_CONFIG_ATTRS 6
 /* Number of entries for diskquota.table_size update SQL */
 #define SQL_MAX_VALUES_NUMBER 1000000
+/* Inital number of entries for hash table in quota_info */
+#define INIT_QUOTA_MAP_ENTRIES (128)
+#define AVG_QUOTA_MAP_ENTRIES (diskquota_max_quotas / (diskquota_max_monitored_databases * NUM_QUOTA_TYPES))
 /* Number of entries for hash table in quota_info */
-#define MAX_QUOTA_MAP_ENTRIES (128 * 1024L)
+#define MAX_QUOTA_MAP_ENTRIES (AVG_QUOTA_MAP_ENTRIES < 1024 ? 1024 : AVG_QUOTA_MAP_ENTRIES)
 
 /* TableSizeEntry macro function */
 /* Use the top bit of totalsize as a flush flag. If this bit is set, the size should be flushed into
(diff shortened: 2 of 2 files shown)


src/diskquota.c
```diff
@@ -76,6 +76,8 @@ int  diskquota_worker_timeout     = 60; /* default timeout is 60 seconds */
 bool diskquota_hardlimit          = false;
 int  diskquota_max_workers        = 10;
 int  diskquota_max_table_segments = 0;
+int  diskquota_max_database       = 0;
+int  diskquota_max_quotas         = 0;
```
fanfuxiaoran, line 80: Maybe init it with default value.

src/quotamodel.c
```diff
@@ -247,6 +252,39 @@ static bool get_table_size_entry_flag(TableSizeEntry *entry, TableSizeEntryFlag
 static void reset_table_size_entry_flag(TableSizeEntry *entry, TableSizeEntryFlag flag);
 static void set_table_size_entry_flag(TableSizeEntry *entry, TableSizeEntryFlag flag);
 
+static struct QuotaMapEntry *
+search_quota_map_entry(HTAB *quota_info_map, struct QuotaMapEntryKey *key, bool *found)
+{
+	struct QuotaMapEntry *entry;
+	uint32                counter = pg_atomic_read_u32(diskquota_quota_info_entry_num);
```
fanfuxiaoran, line 259: Can the `counter` be a bottle when multiple databases run in parallel? And what's more,  the `diskquota_quota_info_entry_num` is not exactly the number of entries allocated in the shared memory, as there is a freelist maintained in hash table. Can we just simply call the `hash_get_num_entries`  to sum the total entries which takes into account freelist and takes no lock?

src/diskquota.c
zhrt123: They will be initialized during GUC definition.

src/quotamodel.c
zhrt123: > Can the counter be a bottle when multiple databases run in parallel
It would not be a bottleneck since the write operation only happens when a new table is created.

> And what's more, the diskquota_quota_info_entry_num is not exactly the number of entries allocated in the shared memory, as there is a freelist maintained in hash table. Can we just simply call the hash_get_num_entries to sum the total entries which takes into account freelist and takes no lock?
This is a crude solution. I added a new PR to solve this problem.

src/quotamodel.c
```diff
@@ -247,6 +254,39 @@ static bool get_table_size_entry_flag(TableSizeEntry *entry, TableSizeEntryFlag
 static void reset_table_size_entry_flag(TableSizeEntry *entry, TableSizeEntryFlag flag);
 static void set_table_size_entry_flag(TableSizeEntry *entry, TableSizeEntryFlag flag);
 
+static struct QuotaMapEntry *
+search_quota_map_entry(HTAB *quota_info_map, struct QuotaMapEntryKey *key, bool *found)
```
Sasasu, line 258: I don't like this function, it is mixing reading and writing paths.
this PR only relates to the write paths. The write path is rarely called than read path. for the write path, it should be user friendly and designed to speed up the read path.

I prefer this logic. this version split the read/write path. while reading the code I can find out what this function is doing without looking inside it.

and the log will not be printed every worker loop (correct me if I am wrong), It should show every times user updates the config (both for file oid update or new quota added)

```c
static struct QuotaMapEntry *
search_quota_map_entry(HTAB *quota_info_map, struct QuotaMapEntryKey *key,HASHACTION action, bool *found) {
  if (action == HASH_FIND) {
    return hash_search(quota_info_map, key, HASH_FIND, found);
  }
  
  if (action == HASH_ENTER) {
    uint32 counter = pg_atomic_read_u32(diskquota_quota_info_entry_num);
    if (counter >= diskquota_max_quotas) {
      /*
       * Too many quotas have been added to the quota_info_map, to avoid diskquota using
       * too much share memory, just return NULL. The diskquota won't work correctly
       * anymore.
       */
      ereport(WARNING, (errmsg("[diskquota] the number of quota exceeds the limit, please increase "
        "the GUC value for diskquota.max_quotas. Current "
        "diskquota.max_quotas value: %d",
        diskquota_max_quotas)));
      return search_quota_map_entry(quota_info_map, key, HASH_FIND, found);
    }
    
    struct QuotaMapEntry * entry = hash_search(quota_info_map, key, HASH_ENTER, found);
    if (!found) {
      counter = pg_atomic_add_fetch_u32(diskquota_quota_info_entry_num, 1);
    }
    
    return entry;
  }
  
  // not implement yet.
  Assert(action == HASH_FIND || action == HASH_ENTER);
}
```c

File: src/quotamodel.c [hunk 4]
```diff
@@ -534,8 +577,12 @@ DiskQuotaShmemSize(void)
 
 	if (IS_QUERY_DISPATCHER())
 	{
+		int num_quota_info_map = diskquota_max_database * NUM_QUOTA_TYPES;
+
 		size = add_size(size, diskquota_launcher_shmem_size());
 		size = add_size(size, sizeof(pg_atomic_uint32));
+		size = add_size(size,
+		                num_quota_info_map * hash_estimate_size(MAX_QUOTA_MAP_ENTRIES, sizeof(struct QuotaMapEntry)));
```
Sasasu, line 585:
```
#define INIT_QUOTA_MAP_ENTRIES (128)
#define AVG_QUOTA_MAP_ENTRIES (diskquota_max_quotas / (diskquota_max_database * NUM_QUOTA_TYPES))
#define MAX_QUOTA_MAP_ENTRIES (AVG_QUOTA_MAP_ENTRIES < 1024 ? 1024 : AVG_QUOTA_MAP_ENTRIES)

int num_quota_info_map = diskquota_max_database * NUM_QUOTA_TYPES;
```

so, the new one

```
num_quota_info_map * hash_estimate_size(MAX_QUOTA_MAP_ENTRIES, sizeof(struct QuotaMapEntry)));
```

 is not equal to the old one

```
hash_estimate_size(MAX_QUOTA_MAP_ENTRIES * NUM_QUOTA_TYPES, sizeof(struct QuotaMapEntry)
```

we're not actually allocating hash tables per database yet, and the hash table size doesn't grow linearly. can we move the `num_quota_info_map` into `hash_estimate_size` to keep the logic correct?

src/quotamodel.c
zhrt123: > and the log will not be printed every worker loop (correct me if I am wrong), It should show every times user updates the config (both for file oid update or new quota added)

It's designed to print log once to avoid too much logs.

zhrt123: This function only uses `HASH_ENTER`. The function name may be inappropriate, we can change it to `put_quota_map_entry`.

src/quotamodel.c
zhrt123: `num_quota_info_map` should not be put in `hash_estimate_size`. Because `num_quota_info_map * hash_estimate_size(MAX_QUOTA_MAP_ENTRIES, sizeof(struct QuotaMapEntry)));` is larger than `hash_estimate_size(MAX_QUOTA_MAP_ENTRIES * NUM_QUOTA_TYPES, sizeof(struct QuotaMapEntry)
`.
`

src/quotamodel.c
```diff
@@ -277,7 +322,9 @@ update_limit_for_quota(int64 limit, float segratio, QuotaType type, Oid *keys)
 		struct QuotaMapEntryKey key = {0};
 		memcpy(key.keys, keys, quota_info[type].num_keys * sizeof(Oid));
 		key.segid                   = i;
-		struct QuotaMapEntry *entry = hash_search(quota_info[type].map, &key, HASH_ENTER, &found);
+		struct QuotaMapEntry *entry = put_quota_map_entry(quota_info[type].map, &key, &found);
+		/* If the number of quota exceeds the limit, entry will be NULL */
```
fanfuxiaoran, line 326: Is it possible for us to merge all the different types of quota config into one hashmap and add a "quota_type" into the `QuotaMapEntryKey` struct? I think the quota will not be too big to fit in one hashmap.

zhrt123: Yes, I also want to merge them, but the change is a little big. I can do it in another PR.

fanfuxiaoran (approved):

src/quotamodel.c
Sasasu: > It's designed to print log once to avoid too much logs.

log every time when user changes the quota setting is not "too much log".

zhrt123: You are right. I will add more information for this in another PR.

PR #373, opened 2023-09-07
Add DiskquotaStatus.

zhrt123:
waiting for #372


src/diskquota_stat.c
```diff
@@ -0,0 +1,191 @@
...
+
+	diskquota_stat = ShmemInitStruct("diskquota status struct", sizeof(DiskquotaStatus), &found);
+	/* initalize diskquota_stat */
+	if (!found)
+	{
+		pg_atomic_init_u32(&diskquota_stat->totale_table_size_entries, 0);
+		pg_atomic_init_u32(&diskquota_stat->totale_quota_info_entries, 0);
+		diskquota_stat->allocated_table_size_entries = ShmemInitStruct(
+		        "allocated_table_size_entries", diskquota_max_database * sizeof(pg_atomic_uint32), NULL);
+		diskquota_stat->used_table_size_entries =
```
fanfuxiaoran, line 61: Why not call `hash_get_num_entries ` to get the allocated items in the hash table?

zhrt123: `hash_get_num_entries` cannot get the number of free entries.

src/diskquota.c
```diff
@@ -89,11 +90,8 @@ static DiskQuotaWorkerEntry *volatile MyWorkerInfo = NULL;
 // how many database diskquota are monitoring on
 static int num_db = 0;
 
-/* how many TableSizeEntry are maintained in all the table_size_map in shared memory*/
-pg_atomic_uint32 *diskquota_table_size_entry_num;
-
-/* how many QuotaMapEntry are maintained in all the quota_info[type].map in shared memory*/
-pg_atomic_uint32 *diskquota_quota_info_entry_num;
+/* The id of current bgworker */
+uint32 current_worker_id;
```
Sasasu, line 94: add `static` to avoid link time conflict

src/diskquota_stat.h
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * -------------------------------------------------------------------------
+ */
+#ifndef DISKQUOTA_STAT_H
+#define DISKQUOTA_STAT_H
+
+#include "c.h"
+#include "port/atomics.h"
+
+typedef struct DiskquotaStatus
```
Sasasu, line 18: this is now a `status`, it is an allocator or a limiter.

the `DiskquotaStatus` can block resource allocation. I prefer to rename this structure.

Sasasu (approved):
looks like you want to detect out-of-limit before PostgreSQL's hash table H_ENTER returns NULL.

I think there will be some memory waste, the memory waste is fine?

src/diskquota_stat.h
```diff
@@ -0,0 +1,46 @@
...
+	                                               in shared memory*/
+	pg_atomic_uint32 *allocated_quota_info_entries;
+	pg_atomic_uint32 *used_quota_info_entries;
+	int               quota_info_entry_nelem_alloc;
+} DiskquotaStatus;
+
+extern void init_diskquota_status(uint32 id);
+
+extern void        decrease_num_table_size_entries(void);
+extern bool        increase_num_table_size_entries(void);
```
Sasasu, line 37: also, `increase` should be `alloc`. this is not just a status

src/diskquota_stat.h
zhrt123: hmm, it is indeed inappropriate. Do you have any suggestions?

zhrt123:
> looks like you want to detect out-of-limit before PostgreSQL's hash table H_ENTER returns NULL.
> 
> I think there will be some memory waste, the memory waste is fine?

We need to keep diskquota not exceeding the memory that diskquota declaimed, although we may waste some memory.

src/diskquota_stat.h
Sasasu: DiskquotaMemoryLimiter or DiskquotaMemoryUsage

zhrt123: We need to add more status to this struct in the next work, like `paused_bgworker_list` showing the bgworkers that are stopped by error. So we'd better use `status` as the struct name.

src/diskquota_stat.h
zhrt123: Increase and decrease are the pair. How about alloc and free?

src/diskquota_stat.c
fanfuxiaoran: 1. `used_table_size_entries` can be replaced by  hash_get_num_entries
2. Record the biggest `used_table_size_entries` as `biggest_table_size_entires`
3. The difference between the `biggest_table_size_entires` and `allocated_table_size_entries` won't be larger than  `nelem_alloc`. So we can use `biggest_table_size_entries` directly as `allocated_table_size_entries` as `nelem_alloc` is utmost to 32, not a big deal. Or you can add`32` to the  `biggest_table_size_entries`

I think this method can significantly simplify the code.

Sasasu (approved):

src/diskquota_stat.c
```diff
@@ -0,0 +1,199 @@
...
+
+/*
+ * Increase the count of memory allocated for TableSizeEntry.
+ * IF used_count = allocated_count, we need to allocate more memory and increase
+ * allocated_count. And allocated_count cannot exceed MAX_NUM_TABLE_SIZE_ENTRIES.
+ */
+bool
+alloc_table_size_entry(void)
+{
+	uint32 used      = diskquota_stat->used_table_size_entries[current_worker_id];
```
fanfuxiaoran, line 104: I still think we should use `hash_get_num_entries`  instead of used_table_size_entries. So you don't need this variable and call `free_quota_info_entry` or increase it.

fanfuxiaoran:
I have a more direct forward idea to collect the shared memory used by one type of hash table. But a little hack
```
HTAB *
ShmemInitHash(const char *name,         /* table string name for shmem index */
                          long init_size,       /* initial table size */
                          long max_size,        /* max size of the table */
                          HASHCTL *infoP,       /* info about key and bucket size */
                          int hash_flags)       /* info about infoP */
{
        bool            found;
        void       *location;

        /*
         * Hash tables allocated in shared memory have a fixed directory; it can't
         * grow or other backends wouldn't be able to find it. So, make sure we
         * make it big enough to start with.
         *
         * The shared memory allocator must be specified too.
         */
        infoP->dsize = infoP->max_dsize = hash_select_dirsize(max_size);
        infoP->alloc = ShmemAllocNoError;
        hash_flags |= HASH_SHARED_MEM | HASH_ALLOC | HASH_DIRSIZE;

        /* look it up in the shmem index */
        location = ShmemInitStruct(name,
                                                           hash_get_shared_size(infoP, hash_flags),
                                                           &found);

        /*
```
For the above function, modify it with the following steps:
1. Create a new function to wrapper ShmemAllocNoError, such as TableSizeShmemAllocNoError, in function `ShmemAllocNoError`, we collect the usages of memory by table_size. 
2. rewrite `ShmemInitHash` to use the new alloc function, for instance, `TableSizeShmemAllocNoError`
Personally, I prefer this idea although I don't like it as it is too hack. This idea is just a suggestion.

src/diskquota_stat.c
```diff
@@ -0,0 +1,199 @@
...
+	total = pg_atomic_add_fetch_u32(&diskquota_stat->total_table_size_entries,
+	                                diskquota_stat->table_size_entry_nelem_alloc);
+	if (total > MAX_NUM_TABLE_SIZE_ENTRIES) return false;
+	/* Increase the allocated count and used count */
+	diskquota_stat->allocated_table_size_entries[current_worker_id] += diskquota_stat->table_size_entry_nelem_alloc;
+	diskquota_stat->used_table_size_entries[current_worker_id]++;
+	return true;
+}
+
+bool
```
fanfuxiaoran, line 124: Looks this function is  duplicate with `alloc_table_size_entry`, can we remove this function and use `alloc_table_size_entry` directly? And report the warning message in `alloc_table_size_entry`

src/diskquota_stat.c
```diff
@@ -0,0 +1,199 @@
...
+{
+	uint32 used      = diskquota_stat->used_quota_info_entries[current_worker_id];
+	uint32 allocated = diskquota_stat->allocated_quota_info_entries[current_worker_id];
+	if (used < allocated) return true;
+	uint32 total = pg_atomic_read_u32(&diskquota_stat->total_quota_info_entries);
+	return total < diskquota_max_quotas;
+}
+
+const char *
+diskquota_status_quota_info_entry(void)
```
fanfuxiaoran, line 192: Is there a related sql function for it? or How it be added to status?

zhrt123: It is used in `diskquota_status(PG_FUNCTION_ARGS)`.

src/diskquota_stat.c
zhrt123: If we remove this function, how to deal this error report:
```
bool exceed_limit = table_size_entry_exceed_limit();
if (exceed_limit) break;
bool success = alloc_table_size_entry();
if (success)
{
	tsentry = (TableSizeEntry *)hash_search(table_size_map, &key, HASH_ENTER, NULL);
        ...
}
/* For the first exceeding, print a warning to the log. */
if (table_size_entry_exceed_limit())
{
	ereport(WARNING, (errmsg("[diskquota] the number of tables exceeds the limit, please increase "
								"the GUC value for diskquota.max_table_segments. Current "
								"diskquota.max_table_segments value: %d",
								diskquota_max_table_segments)));
}
```

PR #380, opened 2023-09-19
Refactor quota info map

zhrt123:
Merge all types of quota info maps into one in each bgworker.

### DIFF
--- a/src/quotamodel.c
+++ b/src/quotamodel.c
@@ -50,16 +50,10 @@
 #define INIT_DISK_QUOTA_REJECT_ENTRIES 8192
 /* per database level max size of rejectmap */
 #define MAX_LOCAL_DISK_QUOTA_REJECT_ENTRIES 8192
-#define MAX_NUM_KEYS_QUOTA_MAP 8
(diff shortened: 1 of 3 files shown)


src/quotamodel.c
```diff
@@ -380,23 +323,24 @@ add_quota_to_rejectmap(QuotaType type, Oid targetOid, Oid tablespaceoid, bool se
  * the quota limit, if it does, add it to the rejectmap.
  */
 static void
-check_quota_map(QuotaType type)
+refresh_quota_info_map(void)
```
fanfuxiaoran, line 326: I thought check_quota_map is better as it will check the limit and add some into the rejectmap

fanfuxiaoran (approved):

src/quotamodel.c
zhrt123: hmm, but in the function, the main procedure is to refresh the quota info map.

Sasasu (approved):
the logic is clearer than before.
And the memory usage is the same as before, due to struct padding.

I think it will be a hard review, but it is not.
nice work.

PR #381, opened 2023-09-26
Add rocky9 to CI

zhrt123:
DEV pipeline: https://dev2.ci.gpdb.pivotal.io/teams/gp-extensions/pipelines/dev.diskquota.add-rocky9.zhrt


concourse/pipeline/job_def.lib.yml
```diff
@@ -228,11 +254,31 @@ plan:
...
     resource: #@ conf["res_diskquota_bin"]
+#@   end
 - #@ _build_task(conf)
 - #@ _test_task(conf)
 - put: #@ conf["res_intermediates_bin"]
   params:
     file: diskquota_artifacts/diskquota.tar.gz
 #@ end
+
+#@ def get_groups(confs):
```
liuxueyang, line 268: There is a bug in this function. Use `get_groups` in base.lib.yml instead.
For example https://github.com/pivotal/gp-extensions-ci/blob/main/projects/advanced-password-check/pr.yml#L23

liuxueyang: If diskquota has not been migrated to the monorepo. I think the only way is to copy the function from https://github.com/pivotal/gp-extensions-ci/blob/main/common/base.lib.yml#L65 manually here.

zhrt123: hmm, I will fix it.

zhrt123:
The other PR has done the same thing, close this PR.

PR #397, opened 2023-10-30
Move quota

zhrt123:
- remove diskquota_quota_info_entry_num
- move quota_info_map to quota.c

### DIFF
--- a/src/diskquota.c
+++ b/src/diskquota.c
@@ -80,9 +80,6 @@ static DiskQuotaWorkerEntry *volatile MyWorkerInfo = NULL;
 // how many database diskquota are monitoring on
 static int num_db = 0;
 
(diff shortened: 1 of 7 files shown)


src/diskquota.c
```diff
@@ -80,9 +80,6 @@ static DiskQuotaWorkerEntry *volatile MyWorkerInfo = NULL;
 // how many database diskquota are monitoring on
 static int num_db = 0;
 
-/* how many QuotaInfoEntry are maintained in all the quota_info_map in shared memory*/
-pg_atomic_uint32 *diskquota_quota_info_entry_num;
```
fanfuxiaoran, line 84: Is this unnecessary now?

zhrt123: YES, we will move the quota_info_map to local memory.

fanfuxiaoran (approved):

PR #399, opened 2023-10-30
Fix memory leak

zhrt123:
StringInfoData should be free in init/vacuum hashtable function.


zhrt123:
> So much changes?

I've rebased the branch with `move-reject-map`. The changes will look less.

fanfuxiaoran (approved):

fanfuxiaoran:
LGTM! Maybe use the memory context to manage the memory is better in the future.

beeender:
Does the leak happen to 2.x as well?

zhrt123:
> Member
> ###
> ** [beeender](/beeender) ** commented [Oct 31, 2023](#issuecomment-1787289669)

No, this is caused by the changes in 3.0.

PR #406, opened 2023-11-22
Invalidate diskquota.table_size entries during startup

RekGRpth:
Invalidate diskquota.table_size entries during startup

Diskquota calculates sizes and stores information in the diskquota.table_size
table periodically with a pause in diskquota.naptime, 2 seconds by default.
If we restart the cluster during this pause, then diskquota will lose all
changes that have occurred since the last save to the diskquota.table_size
table. We could create temporary tables, wait when it will be flushed to
diskquota.table_size table, restart the cluster, and diskquota would remember
the information about the temporary tables. Or we could delete the tables,
restart the cluster, and again diskquota will remember information about the
deleted tables. This happens because at the start of the cluster, diskquota
remembers all the information written to the diskquota.table_size table,
but does not check that some tables may have already been deleted.

As a solution, we invalidate diskquota.table_size during diskquota
worker start in addition to pg_class validation.

### DIFF
--- a/tests/isolation2/sql/test_dropped_table.sql
+++ b/tests/isolation2/sql/test_dropped_table.sql
@@ -20,6 +20,7 @@
 -- Indicates that there is no dropped table in pg_catalog.pg_class
 1: SELECT oid FROM pg_catalog.pg_class WHERE relname = 'dropped_table';
 -- Indicates that there are no entries in diskquota.table_size that are not present in pg_catalog.pg_class
+1: SELECT diskquota.wait_for_worker_new_epoch();
 1: SELECT tableid FROM diskquota.table_size WHERE NOT EXISTS (SELECT 1 FROM pg_catalog.pg_class WHERE tableid = oid) AND segid = -1;
 1: DROP SCHEMA dropped_schema CASCADE;
 1q:
--- a/tests/isolation2/sql/test_dropped_table.sql
+++ b/tests/isolation2/sql/test_dropped_table.sql
@@ -1,0 +1,22 @@
+-- Ensure diskquota does not save information about dropped table during restart cluster by invalidates it at startup
+
+1: CREATE SCHEMA dropped_schema;
+1: SET search_path TO dropped_schema;
+1: SELECT diskquota.set_schema_quota('dropped_schema', '1 MB');
+1: SELECT diskquota.wait_for_worker_new_epoch();
+1: CREATE TABLE dropped_table(id int) DISTRIBUTED BY (id);
+1: INSERT INTO dropped_table SELECT generate_series(1, 100000);
+-- Wait for the diskquota bgworker refreshing the size of 'dropped_table'.
+1: SELECT diskquota.wait_for_worker_new_epoch();
+1: DROP TABLE dropped_table;
+1q:
+
+-- Restart cluster fastly
+!\retcode gpstop -afr;
+
+-- Indicates that there is no dropped table in pg_catalog.pg_class
+1: SELECT oid FROM pg_catalog.pg_class WHERE relname = 'dropped_table';
+-- Indicates that there are no entries in diskquota.table_size that are not present in pg_catalog.pg_class
+1: SELECT tableid FROM diskquota.table_size WHERE NOT EXISTS (SELECT 1 FROM pg_catalog.pg_class WHERE tableid = oid) AND segid = -1;
+1: DROP SCHEMA dropped_schema CASCADE;
+1q:
--- a/tests/isolation2/sql/test_dropped_table.sql
+++ b/tests/isolation2/sql/test_dropped_table.sql
@@ -1,6 +1,6 @@
 -- Ensure diskquota does not save information about dropped table during restart cluster by invalidates it at startup
 
-!\retcode gpconfig -c diskquota.naptime -v 2 --skipvalidation;
+!\retcode gpconfig -c diskquota.naptime -v 5 --skipvalidation;
 !\retcode gpstop -u;
 
 1: CREATE SCHEMA dropped_schema;
@@ -8,7 +8,7 @@
 1: SELECT diskquota.set_schema_quota('dropped_schema', '1 MB');
 1: SELECT diskquota.wait_for_worker_new_epoch();
 1: CREATE TABLE dropped_table(id int) DISTRIBUTED BY (id);
-1: INSERT INTO dropped_table SELECT generate_series(1, 100000);
+1: INSERT INTO dropped_table SELECT generate_series(1, 10000);
 -- Wait for the diskquota bgworker refreshing the size of 'dropped_table'.
 1: SELECT diskquota.wait_for_worker_new_epoch();
 1: DROP TABLE dropped_table;
--- a/tests/isolation2/sql/test_dropped_table.sql
+++ b/tests/isolation2/sql/test_dropped_table.sql
@@ -1,5 +1,8 @@
 -- Ensure diskquota does not save information about dropped table during restart cluster by invalidates it at startup
 
+!\retcode gpconfig -c diskquota.naptime -v 2 --skipvalidation;
+!\retcode gpstop -u;
+
 1: CREATE SCHEMA dropped_schema;
 1: SET search_path TO dropped_schema;
 1: SELECT diskquota.set_schema_quota('dropped_schema', '1 MB');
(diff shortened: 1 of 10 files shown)


zhrt123:
Thanks for your issue. There indeed is a problem during startup. Diskquota won't refresh the table size changes before reaching `diskquota.naptime`. So if the table size is changed and is not flushed into `diskquota.table_size`, diskquota will remember a wrong table size value. We will fix it.

But the example in the PR won't happen in diskquota. Diskquota will check whether a table still exists during every refreshing, so if the table does not exist or is not remembered, diskquota will refresh the information about it.

RekGRpth:
> But the example in the PR won't happen in diskquota. Diskquota will check whether a table still exists during every refreshing, so if the table does not exist or is not remembered, diskquota will refresh the information about it.

No, current isolation2 test sets diskquota.naptime to 0 in config.sql, but if I set diskquota.naptime to 2, then both my new tests are failed without my patch.

RekGRpth:
> Thanks for your issue. There indeed is a problem during startup. Diskquota won't refresh the table size changes before reaching `diskquota.naptime`. So if the table size is changed and is not flushed into `diskquota.table_size`, diskquota will remember a wrong table size value. We will fix it.

How about such diff
```diff
diff --git a/src/diskquota.c b/src/diskquota.c
index 4f0fe43..9e57ff3 100644
--- a/src/diskquota.c
+++ b/src/diskquota.c
@@ -583,7 +583,14 @@ disk_quota_worker_main(Datum main_arg)
 	}
 
 	if (got_sigterm)
+	{
 		ereport(LOG, (errmsg("[diskquota] stop disk quota worker process to monitor database:%s", dbname)));
+		if (!diskquota_is_paused())
+		{
+			refresh_disk_quota_model(MyWorkerInfo->dbEntry);
+			ereport(LOG, (errmsg("[diskquota] refreshed diskquota model before exit")));
+		}
+	}
 	ereport(DEBUG1, (errmsg("[diskquota] stop disk quota worker process to monitor database:%s", dbname)));
 #if DISKQUOTA_DEBUG
 	TimestampDifference(MyWorkerInfo->dbEntry->last_run_time, GetCurrentTimestamp(), &secs, &usecs);
```
But due to the support of [dynamic workers](https://github.com/greenplum-db/diskquota/commit/082ac9a000f2b3ebd803fb69ea8517432aa2148b), this trick will not work for them.

RekGRpth:
I fixed tests reproduction (without patch)

zhrt123:
> ```diff
> refresh_disk_quota_model
> ```

IMHO, the bgworker should not do anything after receiving the sigterm.

zhrt123:
> > Thanks for your issue. There indeed is a problem during startup. Diskquota won't refresh the table size changes before reaching `diskquota.naptime`. So if the table size is changed and is not flushed into `diskquota.table_size`, diskquota will remember a wrong table size value. We will fix it.
> 
> How about such diff
> 
> ```diff
> diff --git a/src/diskquota.c b/src/diskquota.c
> index 4f0fe43..9e57ff3 100644
> --- a/src/diskquota.c
> +++ b/src/diskquota.c
> @@ -583,7 +583,14 @@ disk_quota_worker_main(Datum main_arg)
>  	}
>  
>  	if (got_sigterm)
> +	{
>  		ereport(LOG, (errmsg("[diskquota] stop disk quota worker process to monitor database:%s", dbname)));
> +		if (!diskquota_is_paused())
> +		{
> +			refresh_disk_quota_model(MyWorkerInfo->dbEntry);
> +			ereport(LOG, (errmsg("[diskquota] refreshed diskquota model before exit")));
> +		}
> +	}
>  	ereport(DEBUG1, (errmsg("[diskquota] stop disk quota worker process to monitor database:%s", dbname)));
>  #if DISKQUOTA_DEBUG
>  	TimestampDifference(MyWorkerInfo->dbEntry->last_run_time, GetCurrentTimestamp(), &secs, &usecs);
> ```
> 
> But due to the support of [dynamic workers](https://github.com/greenplum-db/diskquota/commit/082ac9a000f2b3ebd803fb69ea8517432aa2148b), this trick will not work for them.

IMHO, the bgworker should not do anything after receiving the sigterm.

zhrt123:
> > But the example in the PR won't happen in diskquota. Diskquota will check whether a table still exists during every refreshing, so if the table does not exist or is not remembered, diskquota will refresh the information about it.
> 
> No, current isolation2 test sets diskquota.naptime to 0 in config.sql, but if I set diskquota.naptime to 2, then both my new tests are failed without my patch.

You're right. Now there are two problems in diskquota:
- temp tables and deleted tables are still stored in diskquota.table_size.
- incorrect table size cannot be refreshed until the corresponding table becomes an active table.

I'm thinking of a way to solve both of the two problems.

RekGRpth:
> the bgworker should not do anything after receiving the sigterm

[Before output, statistics are saved to disk](https://github.com/greenplum-db/gpdb/blob/22159b4f6b17985fbde2bf478d6431cab0a03a5b/contrib/pg_stat_statements/pg_stat_statements.c#L516-L521)

tests/isolation2/sql/test_temporary_table.sql
```diff
@@ -0,0 +1,27 @@
+-- Ensure diskquota does not save information about temporary table during restart cluster by invalidates it at startup
+
+!\retcode gpconfig -c diskquota.naptime -v 2 --skipvalidation;
```
zhrt123, line 3: ditto

tests/isolation2/sql/test_dropped_table.sql
```diff
@@ -0,0 +1,28 @@
...
+
+!\retcode gpconfig -c diskquota.naptime -v 2 --skipvalidation;
+!\retcode gpstop -u;
+
+1: CREATE SCHEMA dropped_schema;
+1: SET search_path TO dropped_schema;
+1: SELECT diskquota.set_schema_quota('dropped_schema', '1 MB');
+1: SELECT diskquota.wait_for_worker_new_epoch();
+1: CREATE TABLE dropped_table(id int) DISTRIBUTED BY (id);
+1: INSERT INTO dropped_table SELECT generate_series(1, 100000);
```
zhrt123, line 11:
```suggestion
1: INSERT INTO dropped_table SELECT generate_series(1, 10000);
```
To avoid flaky tests, please make table smaller.

tests/isolation2/sql/test_temporary_table.sql
```diff
@@ -0,0 +1,27 @@
...
+!\retcode gpstop -u;
+
+1: CREATE SCHEMA temporary_schema;
+1: SET search_path TO temporary_schema;
+1: SELECT diskquota.set_schema_quota('temporary_schema', '1 MB');
+1: SELECT diskquota.wait_for_worker_new_epoch();
+1: CREATE TEMPORARY TABLE temporary_table(id int) DISTRIBUTED BY (id);
+1: INSERT INTO temporary_table SELECT generate_series(1, 100000);
```
zhrt123, line 11: ditto

zhrt123 (commented):
LGTM, we could merge this PR after applying the following comments. Thanks very much @RekGRpth 

The remaining problem: the incorrect table size cannot be refreshed until the corresponding table becomes an active table.
Solution: call diskquota.init_table_size_table().

tests/isolation2/sql/test_dropped_table.sql
```diff
@@ -0,0 +1,28 @@
+-- Ensure diskquota does not save information about dropped table during restart cluster by invalidates it at startup
+
+!\retcode gpconfig -c diskquota.naptime -v 2 --skipvalidation;
```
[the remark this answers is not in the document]
RekGRpth: applied

tests/isolation2/sql/test_temporary_table.sql
RekGRpth: applied

tests/isolation2/sql/test_temporary_table.sql
RekGRpth: applied

zhrt123: I will merge it after running the CI test. Thanks for your contribution.

zhrt123:
Test results in GP7 are different from those in GP6, I will raise a new PR to fix it. @RekGRpth

RekGRpth:
> Test results in GP7

added expected output for 7X

RekGRpth:
> let's add diskquota.wait_for_worker_new_epoch() before `select tableid`

added

zhrt123 (approved):

PR #407, opened 2023-12-13
Refresh table size map in center worker

zhrt123:
bgworker behavior:
- collect active table size from segments.
- get relation oid list in the current database.
- generate local_table_size_map.
- send database oid, relation oid list and local_table_size_map
  to center worker.

center worker behavior:
- find table_size_map by database oid.
- traverse the relation oid list, and  set TABLE_EXIST flag
  for each entry in table_size_map.
- update table_size_map by local_table_size_map.
- remove the entry without TABLE_EXIST flag.

TODO:
- update quota when the table size is changed.
- update rejectmap in center worker
- send rejectmap to bgworker
- send rejectmap to segments
- replace diskquota.table_size with view and UDF
- when drop extension, vacuum table size map
  in center worker
- do not load table size from diskquota.table_size
  when is_init == true

## Review Notification
- The memory offset calculation mixes (char *) and other type of structs. There may be some wrong.

### DIFF
--- a/src/diskquota_bgworker.c
+++ b/src/diskquota_bgworker.c
@@ -395,8 +395,8 @@ disk_quota_refresh(bool is_init)
 	DiskquotaMessage       *rsp_msg;
 	Size                    msg_sz = 0;
 	ReqMsgRefreshTableSize *req_msg_body;
-	int                     oid_list_length;
-	int                     table_size_map_entry_num;
+	int                     oid_list_len;
+	int                     table_size_entry_list_len;
 
 	// TODO: do not load table size from diskquota.table_size
 	// when is_init == true
@@ -407,8 +407,8 @@ disk_quota_refresh(bool is_init)
 	/* get table size map by local_active_table_map */
 	local_table_size_map = get_current_database_table_size_map(local_active_table_map);
 
-	oid_list_length          = list_length(oidlist);
-	table_size_map_entry_num = hash_get_num_entries(local_table_size_map);
+	oid_list_len              = list_length(oidlist);
+	table_size_entry_list_len = hash_get_num_entries(local_table_size_map);
 
 	/*
 	 * message content:
@@ -417,8 +417,8 @@ disk_quota_refresh(bool is_init)
 	 * - TableSizeEntry list
 	 */
 	msg_sz = add_size(msg_sz, sizeof(ReqMsgRefreshTableSize));
-	msg_sz = add_size(msg_sz, oid_list_length * sizeof(Oid));
-	msg_sz = add_size(msg_sz, table_size_map_entry_num * sizeof(TableSizeEntry));
+	msg_sz = add_size(msg_sz, oid_list_len * sizeof(Oid));
+	msg_sz = add_size(msg_sz, table_size_entry_list_len * sizeof(TableSizeEntry));
 
 	/* attach the meesage looper */
 	looper = attach_message_looper(DISKQUOTA_CENTER_WORKER_MESSAGE_LOOPER_NAME);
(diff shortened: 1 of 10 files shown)


src/diskquota_bgworker.c
```diff
@@ -380,43 +383,71 @@ disk_quota_worker_main_3(Datum main_arg)
...
-	TestMessage *msg_body = (TestMessage *)MSG_BODY(rsp_msg);
-
-	bool ret = rsp_msg->msg_id == req_msg->msg_id && body->a == msg_body->a && body->b == msg_body->b;
-	elog(WARNING, "xxxx %d", ret);
 	free_message(req_msg);
 	free_message(rsp_msg);
-	return NULL;
+	list_free(oidlist);
+	hash_destroy(local_active_table_map);
+	hash_destroy(local_table_size_map);
```
beeender, line 452:
```
void
hash_destroy(HTAB *hashp)
{
	if (hashp != NULL)
	{
		/* allocation method must be one we know how to free, too */
		Assert(hashp->alloc == DynaHashAlloc);
		/* so this hashtable must have its own context */
		Assert(hashp->hcxt != NULL);

		hash_stats("destroy", hashp);

		/*
		 * Free everything by destroying the hash table's memory context.
		 */
		MemoryContextDelete(hashp->hcxt);
	}
}
```

Since the memory context has not been set specifically before hash map creation, I doubt the `hash_destroy` here will cause memory issues.

src/diskquota_center_worker.c
```diff
@@ -81,6 +92,18 @@ diskquota_center_worker_shmem_size(void)
...
 }
 
+static void
+init_main_map(void)
+{
+	HASHCTL ctl;
+	memset(&ctl, 0, sizeof(ctl));
+	ctl.keysize   = NAMEDATALEN;
+	ctl.entrysize = sizeof(HashMap);
+	ctl.hcxt      = CurrentMemoryContext;
```
beeender, line 102: Just use top? This map will never be destroyed i think?

src/diskquota_center_worker.c
```diff
@@ -25,6 +25,16 @@
...
 #include "diskquota_center_worker.h"
 #include "message_def.h"
+#include "table_size.h"
+#include "quota.h"
+
+static HTAB *main_map;
+
+typedef struct HashMap
+{
+	char  name[NAMEDATALEN];
```
beeender, line 35: Is the key the database name? Shall we handle the database renaming?

src/msg_looper.h
```diff
@@ -68,5 +68,9 @@ extern DiskquotaMessage *send_request_and_wait(DiskquotaLooper *looper, Diskquot
 /* message function */
 extern DiskquotaMessage *init_message(DiskquotaMessageID msg_id, size_t payload_len);
 extern void              free_message(DiskquotaMessage *msg);
+extern void              fill_message_content_by_list(void *addr, List *list, Size sz);
+extern void              fill_message_content_by_hash_table(void *addr, HTAB *ht, Size sz);
+extern inline void       copy_value_from_message_content_list(void *list, void *dst, int idx, Size sz);
```
beeender, line 73: Without function body, this inline function may not be "inlined" for `diskquota_center_worker.c` (I am not quite sure if the modern compiler supports this nowadays)

src/diskquota_center_worker.c
```diff
@@ -189,3 +218,148 @@ disk_quota_message_handler(DiskquotaMessage *req_msg)
...
+	table_size_map = hash_search(main_map, hash_map_name.data, HASH_ENTER, &found);
+	if (!found)
+	{
+		table_size_map->map = create_table_size_map(hash_map_name.data);
+	}
+
+	/* set the TABLE_EXIST flag for existing relations. */
+	for (int i = 0; i < oid_list_length; i++)
+	{
+		copy_value_from_message_content_list(req_msg_body->oid_list, &key.reloid, i, sizeof(Oid));
```
beeender, line 265: Does this work? The `*oid_list` in the DSM will be mapped to the **same address** on different processes?

src/table_size.c
```diff
@@ -12,62 +12,133 @@
...
+	memset(&ctl, 0, sizeof(ctl));
+	ctl.keysize    = sizeof(TableSizeEntryKey);
+	ctl.entrysize  = sizeof(TableSizeEntry);
+	ctl.hcxt       = CurrentMemoryContext;
+	table_size_map = diskquota_hash_create(name, 1024, &ctl, HASH_ELEM | HASH_CONTEXT, DISKQUOTA_TAG_HASH);
+	return table_size_map;
 }
 
+// TODO: when drop extension, vacuum table size map
+// in center worker
```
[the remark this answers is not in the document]
zhrt123: I think we just need to call hash_destroy() here without any special memory context. When creating hash table, the sub-memory context `dynahash` is created in the certain memory context, TopMemoryContext in our scene.

src/diskquota_bgworker.c
```diff
@@ -380,43 +383,71 @@ disk_quota_worker_main_3(Datum main_arg)
...
+	int                     table_size_map_entry_num;
+
+	// TODO: do not load table size from diskquota.table_size
+	// when is_init == true
+	/* get active table size from all segments */
+	local_active_table_map = gp_fetch_active_tables(is_init);
+	/* get all relation oids in the current database */
+	oidlist = get_current_database_oid_list();
+	/* get table size map by local_active_table_map */
+	local_table_size_map = get_current_database_table_size_map(local_active_table_map);
```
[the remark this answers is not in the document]
zhrt123: > get_current_database_table_size_map() is questionable. Why do we convert it to table_size_map then convert it to a table_size_request? Why don't we directly convert the active_table to the request to save some CPU and memory?

get_current_database_table_size_map() checks whether a table exists in the current database in the current refresh term. This should be finished in the bgworker. 

And we indeed can just send the active_table to the center worker from bgworker, but the calculation in `get_current_database_table_size_map()` also needs to be done in the center worker.

src/diskquota_bgworker.c
zhrt123: The default memory context is TopMemoryContext.

src/diskquota_center_worker.c
zhrt123: No, the key is the database oid.

src/diskquota_center_worker.c
zhrt123: hmmm, your concern is right, we cannot use `*oid_list` directly.

src/diskquota_bgworker.c
beeender: See implementation of `hash_create()`, a new context will be created there.

beeender (approved):

PR #409, opened 2024-01-04
Add main hash map in center worker

zhrt123:
main hash map maintains the following maps:
- table size map
- quota info map
- reject map


yihong0618:
seems pr pipeline did not trigger?

zhrt123:
> seems pr pipeline did not trigger?

The code is incomplete, it won't pass the CI..

src/main_hash_map.h
```diff
@@ -0,0 +1,22 @@
+#ifndef MAIN_HASH_MAP
+#define MAIN HASH_MAP
+
+#define TableSizeMapNamePrefix "TableSizeMap"
+#define QuotaInfoMapNamePrefix "QuotaInfoMap"
```
beeender, line 5: Add a `DQ` prefix to avoid possible conflicts with others. maybe

```
#define TableSizeMapNamePrefix "DQTableSize"
#define QuotaInfoMapNamePrefix "DQQuotaInfo"
```

zhrt123: What DQ means?

beeender (approved):

PR #412, opened 2024-01-08
Fix bug: the address of message cannot be access in another endpoint.

zhrt123:
While a message is created in client, the address is stored as the virtual memory address in client process. But server cannot access the address. So we need to store the address offset of the message content.

### DIFF
--- a/src/msg_looper.h
+++ b/src/msg_looper.h
@@ -17,18 +17,41 @@
 #include "postgres.h"
 #include "storage/dsm.h"
 
+/*
+ * Example of message struct:
+ * typedef struct ReqMsgSample {
+ * 		int a;
+ * 		int list1_len;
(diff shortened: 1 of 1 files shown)


src/msg_looper.h
```diff
@@ -17,18 +17,41 @@
...
 #include "storage/dsm.h"
 
+/*
+ * Example of message struct:
+ * typedef struct ReqMsgSample {
+ * 		int a;
+ * 		int list1_len;
+ * 		int list2_len;
+ * 		uint64 list1_offset;
+ * 		uint64 list2_offset;
```
beeender, line 27: use `size_t` instead of `int`/`uint64`.

https://en.cppreference.com/w/c/types/offsetof

beeender (approved):

PR #413, opened 2024-01-09
Limit diskquota hash table's size according initial request

KnightMurloc:
Diskquota did not control the size of its hash tables in shared memory and could
have consumed shared memory not intended for it, potentially impacting the other
database subsystems. Hash tables can also grow indefinitely if the background
process on the coordinator has not started, which can happen for a number of
reasons: gone done with error, pause, isn’t started. In this case, data is not
collected from segments and some hash tables (active_tables_map, relation_cache,
relid_cache) would not be cleared and would overflow.

This patch adds a limit on the size of all hash tables in shared memory by
adding a function that checks whether the hash table is full. The function
returns HASH_FIND if the map is full and HASH_ENTER otherwise. It also report a
warning if the table is full. Implemented a GUC that controls how frequently the
warning will be reported, as it could be reported too frequently. Also added a
GUC to control size of local reject map. The size of global reject map is set
as diskquota_max_local_reject_entries * diskquota_max_monitored_databases.

The test_active_table_limit test has been changed. Firstly, the value of
max_active_tables was changed from 2 to 5, since tables from all databases
processed by diskquota are placed in active_tables_map and with a limit of 2
tables overflow occurs even when the extension is created. Secondly, now a
table with 10 partitions is created to overflow active_tables_map, after which
another table is created into which data is inserted that should exhaust the
quota, but since this table does not inserted into active_tables_map, its size
is not taken into account and we can insert into the table after that. At the
end, vacuum full is done to achieve the overflow of altered_reloid_cache.

### DIFF
--- a/src/diskquota.c
+++ b/src/diskquota.c
@@ -70,14 +70,16 @@ static volatile sig_atomic_t got_sigusr1 = false;
 static volatile sig_atomic_t got_sigusr2 = false;
 
 /* GUC variables */
-int  diskquota_naptime                 = 0;
-int  diskquota_max_active_tables       = 0;
-int  diskquota_worker_timeout          = 60; /* default timeout is 60 seconds */
-bool diskquota_hardlimit               = false;
-int  diskquota_max_workers             = 10;
-int  diskquota_max_table_segments      = 0;
-int  diskquota_max_monitored_databases = 0;
-int  diskquota_max_quota_probes        = 0;
+int  diskquota_naptime                         = 0;
+int  diskquota_max_active_tables               = 0;
+int  diskquota_worker_timeout                  = 60; /* default timeout is 60 seconds */
+bool diskquota_hardlimit                       = false;
+int  diskquota_max_workers                     = 10;
+int  diskquota_max_table_segments              = 0;
+int  diskquota_max_monitored_databases         = 0;
+int  diskquota_max_quota_probes                = 0;
+int  diskquota_max_local_reject_entries        = 0;
+int  diskquota_hashmap_overflow_report_timeout = 0;
 
 DiskQuotaLocks       diskquota_locks;
 ExtensionDDLMessage *extension_ddl_message = NULL;
@@ -88,12 +90,6 @@ static DiskQuotaWorkerEntry *volatile MyWorkerInfo = NULL;
 
 // how many database diskquota are monitoring on
 static int num_db = 0;
-
-/* how many TableSizeEntry are maintained in all the table_size_map in shared memory*/
-pg_atomic_uint32 *diskquota_table_size_entry_num;
-
-/* how many QuotaInfoEntry are maintained in all the quota_info_map in shared memory*/
-pg_atomic_uint32 *diskquota_quota_info_entry_num;
 
 static DiskquotaLauncherShmemStruct *DiskquotaLauncherShmem;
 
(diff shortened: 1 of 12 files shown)


zhrt123:
The idea is very nice, but there are some problems with this modification:
- diskquota_max_quota_probes is the limit of quota_info_map in the entire cluster, we cannot use it to limit the quota_info_map in one bgworker. (every bgworker uses different quota_info_map).
- MAX_NUM_TABLE_SIZE_ENTRIES is the same as diskquota_max_quota_probes.
- The limit for active_table_map, rejectmap and altered_reloid_cache is set to 1000000 by default, which is large enough for most clusters. The limit also can be set by diskquota.max_active_tables.

KnightMurloc:
> MAX_NUM_TABLE_SIZE_ENTRIES is the same as diskquota_max_quota_probes.

To limit table_size_map, I use the same macro as when creating the table.

https://github.com/arenadata/diskquota/blob/4095f4be4487b00b8931360b7734a1cd8478d568/src/quotamodel.c#L999-L1000

https://github.com/greenplum-db/diskquota/blob/9fd7a2232f25ce8810c6df57c8bca9fd3da5549d/src/quotamodel.c#L572-L573

> The limit for active_table_map, rejectmap and altered_reloid_cache is set to 1000000 by default, which is large enough for most clusters. The limit also can be set by diskquota.max_active_tables.

Are you suggesting using max_active_tables to limit reject maps as well? Instead of adding another GUC.

zhrt123:
> > MAX_NUM_TABLE_SIZE_ENTRIES is the same as diskquota_max_quota_probes.
> 
> To limit table_size_map, I use the same macro as when creating the table.
> 
> https://github.com/arenadata/diskquota/blob/4095f4be4487b00b8931360b7734a1cd8478d568/src/quotamodel.c#L999-L1000
> 
> https://github.com/greenplum-db/diskquota/blob/9fd7a2232f25ce8810c6df57c8bca9fd3da5549d/src/quotamodel.c#L572-L573
> 
> > The limit for active_table_map, rejectmap and altered_reloid_cache is set to 1000000 by default, which is large enough for most clusters. The limit also can be set by diskquota.max_active_tables.
> 
> Are you suggesting using max_active_tables to limit reject maps as well? Instead of adding another GUC.

My wrong description. I mean there is the same problem with MAX_NUM_TABLE_SIZE_ENTRIES (MAX_NUM_TABLE_SIZE_ENTRIES is the limit of table_size_map in the entire cluster, we cannot use it to limit the table_size_map in one bgworker, since every bgworker uses different table_size_map.).

> Are you suggesting using max_active_tables to limit reject maps as well? Instead of adding another GUC.

YES.

zhrt123:
We need to limit the number of hash table entries in the whole cluster. For example,
we set the limit for the number of TableSizeEntry to 1000.
- db1 has 800 entries
- db2 has 250 entries
The number of entries exceeds the limit.

But in this PR, it won't be considered as exceeding the limit, because the number of entries in db1 and db2 does not exceed 1000.

KnightMurloc:
Okay, I get it. Are there any reasons why we want to limit this map per cluster and not per database? Also, as I understand it, such logic is assumed only for table_size_map?

zhrt123:
> Are there any reasons why we want to limit this map per cluster and not per database?

Because the shared memory is allocated for the whole cluster, the related GUC is used to limit the shared memory usage.

> Also, as I understand it, such logic is assumed only for table_size_map?

No, it also suits `quota_info_map`.

RekGRpth:
How about, instead of using two atomics, make those two hashmaps for the entire cluster, and not for each database?

PR #414, opened 2024-01-09
Free tuple after use in refresh_rejectmap

dnskvlnk:
refresh_rejectmap looks for a tuple using SearchSysCacheCopy1 which retrieves a copy of the tuple allocating memory for it. However, refresh_rejectmap didn't free these tuple copies after use. If lots of oids were passed, diskquota could work incorrectly because of huge memory leak. This patch frees these tuples and prevents memory leaks.


zhrt123 (approved):
LGTM. Thanks for your contribution. These tuples will be free after UDF is completed. But they still occupy the memory during the UDF is executed. So it is nice for me to free the tuple after using it.

